PR #64, opened 2019-03-08
(maint) Do not use in place replacements

ekohl:
Since the new value is already assigned to the variable, there's no need to replace the string in place.


tphoney:
thanks for the PR @ekohl 👍  great work

PR #65, opened 2019-03-08
Introduce a litmus:metadata task

ekohl:
This task prints the supported OSes from metadata. It also splits off the parsing of metadata into a separate function so that could be tested standalone.

### DIFF
--- a/lib/puppet_litmus/rake_tasks.rb
+++ b/lib/puppet_litmus/rake_tasks.rb
@@ -8,6 +8,42 @@ require 'pdk'
 require 'json'
 require 'parallel'
 
+def get_metadata_operating_systems(metadata)
+  return unless metadata.is_a?(Hash)
+  return unless metadata['operatingsystem_support'].is_a?(Array)
+
+  metadata['operatingsystem_support'].each do |os_info|
+    next unless os_info['operatingsystem'] && os_info['operatingsystemrelease']
+
+    os_name = case os_info['operatingsystem']
+              when 'Amazon', 'Archlinux', 'AIX', 'OSX'
+                next
+              when 'OracleLinux'
+                'oracle'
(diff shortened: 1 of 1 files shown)


tphoney:
great work @ekohl , the mapping of names from metadata to image names. EG docker or vagrant. is not easy. I have ideas 🏋️. Specifically i think the metadata file should contain sections for different kinds of testing. These sections are not used by the forge. 
it would mean that we can explicitly list out the actual os image names we would expect to be provisioned. rather than have a mappings function.

ekohl:
Changing the section would mean changes to https://github.com/mcanevet/rspec-puppet-facts. It might also be good to exactly match the names used to what rspec-puppet-facts does and then have a separate translation layer to images. Images may also depend on the virtualization layer that's used. For example, https://app.vagrantup.com/ubuntu only provides images for the virtualbox provider but https://app.vagrantup.com/generic/ does provide images for the libvirt provider. When using openstack you can also use a regular expression. https://github.com/theforeman/forklift/tree/master/vagrant/boxes.d is what we came up with in our testing setup.

Related to this PR, I have thought about supporting SPEC_FACTS_OS following https://github.com/mcanevet/rspec-puppet-facts#running-your-tests.

tphoney:
Great PR, tested locally it works as expected. i wont merge until monday when we can create an issue / ticket or something to capture what we have talked about here. potentially have a call to talk through what we all want. Great work @ekohl !

👍

ekohl:
Looks like I already did something similar in Voxpupuli: https://github.com/voxpupuli/modulesync_config/blob/626a2e6f9cf39eed6b33e6426f9758026b76a425/moduleroot/Rakefile.erb#L80-L99 (but didn't update it to provide a `beaker-hostgenerator` compatible output.

Issue #66, opened 2019-03-08
Declare PDK as a dependency


ekohl:
a1a3c6701a43394d71bd74ae571931045728bef1 removed PDK as a dependency which means you can't just include puppet_litmus in your Gemfile. Is the pinning still needed?

pmcmaw:
Hey @ekohl

This is an issue we are aware of and currently working on it. There was a release of bolt last night to update version of cri, this means we can update the version of cri in the pdk which will resolve our dependency issue that we have previously ran into. This will be resolved within the next few days.

Also it is fantastic to see movement on this repo by you :-) feedback etc has been really appreciated and something that we are going to heavily rely on to ensure we are creating one solution for everyone :-)

trevor-vaughan:
I don't think that PDK should be a dependency.

People have different frameworks and test in different ways. PDK may be a useful method for getting there for new users but should not be a hard requirement since it makes it more difficult for existing organizations to migrate.

ekohl:
@tavilarion I personally agree. Now that Puppet 6 can no longer build modules I would like a very light gem that just implements the module building. PDK can depend on this to provide the same functionality. Litmus could also depend on just this gem. [puppet-blacksmith](https://github.com/voxpupuli/puppet-blacksmith) would add the dependency as well.

tphoney:
Hi all, i have created [this](https://github.com/puppetlabs/puppet_litmus/blob/master/README.md#known-issues) which details what is going on. We are working to get this work merged, so things become more transparent and obvious. The ruby ecosystem 😞

PR #67, opened 2019-03-08
Require bolt >= 0.21.4 and remove pry requirement

ekohl:
This bolt version refactored and introduced bolt_spec/run. Pry shouldn't be a requirement at runtime.

https://github.com/puppetlabs/puppet_litmus/issues/66 is still a problem with this.

### DIFF
--- a/lib/puppet_litmus.rb
+++ b/lib/puppet_litmus.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require 'pry'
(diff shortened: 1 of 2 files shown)


tphoney:
many thanks @ekohl , i have been busy the past few days with other team commitments. i am working in the background to get the bolt and pdk gem to use the same cri version. this will allow us to have both dependencies in the gemspec. 
simplifying things

PR #85, opened 2019-04-12
(maint) move inventory into separate module

tphoney:


lib/puppet_litmus.rb
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
 require 'bolt_spec/run'
+require 'inventory'
```
ekohl: Should this be `puppet_litmus/inventory` and the filename of it `lib/puppet_litmus/inventory.rb`?

tphoney:
thanks @ekohl i made the changes. i will likely move the apply_manifest functions out  next.

lib/inventory_manipulation.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+# helper functions for manipulating and reading a bolt inventory file
+module InventoryManipulation
```
ekohl: Should it be namespaced so it ends up as `Litmus::InventoryManipulation`?

tphoney: not 100% sure. TBH. It would be more practical long term for this to live in the bolt gem, leaving it as a simple module would aid this. 
However the full name does make it clear where the functionality is coming from.

lib/inventory_manipulation.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+    targets
+  end
+
+  def target_in_group(inventory_hash, node_name, group_name)
+    exists = false
+    inventory_hash['groups'].each do |group|
+      next unless group['name'] == group_name
+
+      group['nodes'].each do |node|
```
ekohl: Probably not the right time to do this, but:
```ruby
def target_in_group(inventory_hash, node_name, group_name)
  inventory_hash['groups'].any? do |group|
    group['name'] == group_name && group['nodes'].any? { node['name'] == node_name }
  end
end
```

Also, what's the whole pattern where it's a list with an entry name in it? `metadata.json` also has this madness. What's wrong with a hash? Then it's a simple `group = inventory_hash['groups'][group_name] ; !group.nil? && group.include?(node_name)`. Can probably also use `.try()` after it's dropped Puppet 4/Ruby 2.1.

tphoney: i'll have a look :) it may not go in this PR.

lib/inventory_manipulation.rb
tphoney: namespaced it

lib/inventory_manipulation.rb
ekohl: Probably best not to do it in this PR :)

lib/inventory_manipulation.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
```
dhollinger: Shouldn't this file go in `lib/puppet_litmus/inventory_manipulation.rb` to follow Ruby module folder structure convention?

tphoney:
@ekohl moved and i squashed the commits. i will be moving the spec helper functions out into their own module next.

PR #90, opened 2019-04-18
(FM-7891) separate serverspec helpers and add unit

tphoney:


lib/puppet_litmus/serverspec_helpers.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+# helper functions for running puppet commands, and helpers
+module PuppetLitmus::ServerspecHelpers
```
ekohl: Going to do some bikeshedding: `PuppetLitmus::Helpers::Serverspec` perhaps?

tphoney: moved it to PuppetLitmus::Serverspec

florindragos:
👍

PR #92, opened 2019-04-25
(FM-7718) Support vagrant provisioning

florindragos:


lib/puppet_litmus/rake_tasks.rb
```diff
@@ -120,8 +120,18 @@ def run_local_command(command)
           puts "#{result['result']['node_name']}, #{args[:platform]}"
         end
       end
+    elsif args[:provisioner] == 'vagrant'
```
ekohl: This block is exactly the same as abs and vmpooler. Looks like you can also include `vagrant` on line 103. For that matter, docker could also be added there.

florindragos: That's true. good catch! Started out with the same code thinking it will end up being different...

lib/puppet_litmus/rake_tasks.rb
```diff
@@ -100,7 +100,7 @@ def run_local_command(command)
     config_data = { 'modulepath' => File.join(Dir.pwd, 'spec', 'fixtures', 'modules') }
     raise "the provision module was not found in #{config_data['modulepath']}, please amend the .fixtures.yml file" unless File.directory?(File.join(config_data['modulepath'], 'provision'))
 
-    if %w[abs vmpooler].include?(args[:provisioner])
+    if %w[abs vmpooler docker vagrant].include?(args[:provisioner])
```
ekohl: Perhaps it should be an early raise? Then dedent the rest of the block. I'm also wondering if supported provisioners should be a constant to avoid the duplication.

```suggestion
    unless %w[abs vmpooler docker vagrant].include?(args[:provisioner])
      raise "Unknown provisioner '#{args[:provisioner]}', try abs/docker/vmpooler/vagrant"
    end
```

PR #93, opened 2019-04-25
(FM-7981) raise in provision_list if one fails

tphoney:


lib/puppet_litmus/rake_tasks.rb
```diff
@@ -87,10 +87,20 @@ def run_local_command(command)
   task :provision_list, [:key] do |_task, args|
     provision_hash = YAML.load_file('./provision.yaml')
     provisioner = provision_hash[args[:key]]['provisioner']
+    failed_image_message = ''
```
ekohl: Should it be a hash of image -> error? Makes it easy to separate the formatting and allows you to provide a sorted list.

tphoney: @ekohl that was the way i had it for my first attempt. It didnt really add much, and this just simplifies the code. We cant use a hash, because there may be more than one entry for a given image. EG i want to spin up 3 centos-7 machines. if in a hash the errors would be masked by each other.

ekohl: Fair enough

lib/puppet_litmus/rake_tasks.rb
```diff
@@ -87,10 +87,20 @@ def run_local_command(command)
...
+      $stdout = capture_rake_output
       Rake::Task['litmus:provision'].invoke(provisioner, image)
+      if $stdout.string =~ %r{.status.=>.failure}
+        failed_image_message += "=====\n#{image}\n#{$stdout.string}\n"
+      else
+        STDOUT.puts $stdout.string
+      end
       Rake::Task['litmus:provision'].reenable
     end
+    raise "Failed to provision with '#{provisioner}'\n #{failed_image_message}" unless failed_image_message == ''
```
ekohl:
```suggestion
    raise "Failed to provision with '#{provisioner}'\n #{failed_image_message}" unless failed_image_message.empty?
```

tphoney:
many thanks @ekohl , changed.

pmcmaw:
Merging :-) LGTM.

PR #99, opened 2019-04-29
(FM-7963) Yard doc for serverspec

tphoney:


ghoneycutt:
I don't see a rake task for this. How do the docs get generated?

tphoney:
working on it :)

tphoney:
image
Its a start, inventory needs to be documented. So do the rake tasks. And outputting to markdown would mean we could update the wiki.
Its a start.

lib/puppet_litmus/serverspec.rb
```diff
@@ -1,13 +1,25 @@
...
     manifest_file_location = create_manifest_file(manifest)
     apply_manifest(nil, catch_failures: true, manifest_file_location: manifest_file_location)
     apply_manifest(nil, catch_changes: true, manifest_file_location: manifest_file_location)
   end
 
+  # Applies a manifest. returning the result of that apply. Mimics the apply_manifest from beaker
+  #
+  # @param manifest [String] puppet manifest code to be applied.
+  # @param opts [Hash] Alters the behaviour of the command. Valid options are:
+  # :catch_changes [Boolean] exit status of 1 if there were changes.
```
ekohl: From the rendered version it looks like this isn't caught. Probably needs indenting.

tphoney:
comments are a pain, rubocop and yard fighting. 
image
It works now

lib/puppet_litmus/serverspec.rb
```diff
@@ -1,14 +1,29 @@
...
+  #
+  # @param manifest [String] puppet manifest code to be applied.
+  # @return [Boolean] The result of the 2 apply manifests.
   def idempotent_apply(manifest)
     manifest_file_location = create_manifest_file(manifest)
     apply_manifest(nil, catch_failures: true, manifest_file_location: manifest_file_location)
     apply_manifest(nil, catch_changes: true, manifest_file_location: manifest_file_location)
   end
 
+  # rubocop:disable Layout/TrailingWhitespace
```
ekohl: Is the trailing whitespace needed? I think that in Puppet Strings (which is also Yard based) it isn't needed. Or is it to force a newline?

lib/puppet_litmus/serverspec.rb
```diff
@@ -1,14 +1,29 @@
...
+
+  # Applies a manifest. returning the result of that apply. Mimics the apply_manifest from beaker
+  #
+  # @param manifest [String] puppet manifest code to be applied.
+  # @param opts [Hash] Alters the behaviour of the command. Valid options are:
```
ekohl: Long term it'd be nice to change the API to explicitly accept these parameters rather than squashing them in opts, but let's at least document them :)

lib/puppet_litmus/serverspec.rb
tphoney: it is to force the newline, to make it easier to read.

lib/puppet_litmus/serverspec.rb
tphoney: It was to provide some backwards compatibility for the beaker helper. i agree with you

PR #101, opened 2019-05-03
(FM-7978) use a spinner for parallel acceptance

tphoney:


tphoney:
image
this is basic for now, but gives the screen something to do, preventing travis timeouts

Dan33l:
I don't know if this is the same display used here that with pdk but in console of gitlab runners, pdk is hard to be used. I filled a bug about this here:
https://gitlab.com/gitlab-org/gitlab-runner/issues/3570

tphoney:
@Dan33l  thanks for the comments, and yes TTY is hard. i am changing it so there is a travis / pipelines friendly progress bar, that keeps the terminal alive. github closes after 10 minutes if there is no activity. 
@ekohl agreed, what functions were you talking about specifically. 

i thought this would be easy, but working until midnight to get something simple working. proved otherwise

florindragos:
👍

PR #122, opened 2019-06-03
FM-8119 - provision_from_metadata should be deprecated

lionce:


lionce:
hello @ekohl ,
For now bundle exec litmus:provision_from_metadata[provisioner] is working just for vmpooler provider and because of that we're marking it as deprecated. It's using the old behaviour as beaker's provisioning. This PR is just to mark the task as being deprecated. I'll remove the added suggestion and will come back with a better solution for provisioning. I'll create a different ticket for this suggestion. 
Thank you!

PR #126, opened 2019-06-05
(FM-7721) add unit tests for the major parts

tphoney:


spec/lib/puppet_litmus/inventory_manipulation_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    it 'no matching node, raises' do
+      expect { dummy_class.config_from_node(config_hash, 'not.here') }.to raise_error('No config was found for not.here')
+    end
+
+    it 'no config section, returns nil' do
+      expect(dummy_class.config_from_node(no_config_hash, 'test.delivery.puppetlabs.net')).to eq(nil)
+    end
+
+    it 'config exists, and returns' do
+      expect(dummy_class.config_from_node(config_hash, 'test.delivery.puppetlabs.net')).to eq('transport' => 'ssh', 'ssh' =>
```
ThoughtCrhyme: Perhaps we could use string interpolation for the config and the eq matcher just so we don't have to repeat ourselves by writing a matching hash.

tphoney: if it was code, i would agree. When writing tests i prefer to be explicit. Even though it is repeated. When a test fails somewhere strange. plain strings always work, with interpolation you can end up with a nast variable name.

pmcmaw:
LGTM 👍 
Its exciting that tests are going in! 🥇

PR #128, opened 2019-06-06
(feat) add provision_and_install task

tphoney:


lib/puppet_litmus/rake_tasks.rb
```diff
@@ -171,7 +171,7 @@ def run_local_command(command)
     PE_RELEASE = 2019.0
     pe_latest_cmd = "curl http://enterprise.delivery.puppetlabs.net/#{PE_RELEASE}/ci-ready/LATEST"
     pe_latest = run_command(pe_latest_cmd, target_nodes, config: config_data, inventory: inventory_hash)
-    pe_latest_string = pe_latest[0]['result']['stdout'].gsub("\n", '')
+    pe_latest_string = pe_latest[0]['result']['stdout'].delete("\n")
```
eimlav: Would the `.strip` string function not be more suitable here?

tphoney: it was recommended by rubocop. so i'll go with it.

PR #130, opened 2019-06-10
(FM-8185) Use spinner when provisioning 

florindragos:
also check for nodes before running anything on them

### DIFF
--- a/lib/puppet_litmus/rake_tasks.rb
+++ b/lib/puppet_litmus/rake_tasks.rb
@@ -111,6 +111,12 @@ namespace :litmus do
   desc "provision container/VM - abs/docker/vagrant/vmpooler eg 'bundle exec rake 'litmus:provision[vmpooler, ubuntu-1604-x86_64]'"
   task :provision, [:provisioner, :platform] do |_task, args|
     include BoltSpec::Run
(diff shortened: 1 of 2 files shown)


lib/puppet_litmus/rake_tasks.rb
```diff
@@ -111,27 +111,43 @@ def run_local_command(command)
...
     unless %w[abs docker docker_exp vagrant vmpooler].include?(args[:provisioner])
       raise "Unknown provisioner '#{args[:provisioner]}', try abs/docker/vagrant/vmpooler"
     end
-
     params = { 'action' => 'provision', 'platform' => args[:platform], 'inventory' => Dir.pwd }
     results = run_task("provision::#{args[:provisioner]}", 'localhost', params, config: config_data, inventory: nil)
-    raise "Failed provisioning #{args[:platform]} using #{args[:provisioner]}\n#{results.first}" if results.first['status'] != 'success'
 
+    if results.first['status'] != 'success'    
+      puts "Failed provisioning #{args[:platform]} using #{args[:provisioner]}\n#{results.first}"
```
tphoney: this should be a raise. rather than an exit and a puts.

tphoney: at least one unit test would be good. for provision

PR #134, opened 2019-06-12
(FM-8249) add localhost check to run_shell

ThoughtCrhyme:


tphoney:
@ThoughtCrhyme a unit test would be needed for this, even if it is just happy path. following the other tests there.

lib/puppet_litmus/serverspec.rb
```diff
@@ -3,6 +3,8 @@
 # helper functions for running puppet commands. They execute a target system specified by ENV['TARGET_HOST']
 # heavily uses functions from here https://github.com/puppetlabs/bolt/blob/master/developer-docs/bolt_spec-run.md
 module PuppetLitmus::Serverspec
+  @target_node_name = ENV['TARGET_HOST']
```
ThoughtCrhyme: Should probably reference this variable in all the other methods instead of always assigning a local variable.

tphoney: i would prefer this in as a separate change, for now. I worry about it being set once, on include of the library. and it never being updated. possible cross talk in parallel threads. that may be unfounded, but testing it out would require time.

PR #136, opened 2019-06-17
(FM-8268) adding file/directory upload method

tphoney:


tphoney:
A review from @ThoughtCrhyme @florindragos or @michaeltlombardi  would be cool 😺

lib/puppet_litmus/serverspec.rb
```diff
@@ -102,6 +102,33 @@ def run_shell(command_to_run, opts = {})
...
 
+  # Copies file to the target, using its respective transport
+  #
+  # @param source [String] place locally, to copy from.
+  # @param destination [String] place on the target, to copy to.
+  # @param opts [Hash] Alters the behaviour of the command. Valid options are :expect_failures [Boolean] doesnt return an exit code of non-zero if the command failed.
+  # @yieldreturn [Block] this method will yield to a block of code passed by the caller; this can be used for additional validation, etc.
+  # @return [Object] A result object from the command.
+  def bolt_upload_file(source, destination, options = {})
+    inventory_hash = inventory_hash_from_inventory_file
```
ThoughtCrhyme: This will fail if TARGET_HOST is 'localhost' at this line https://github.com/puppetlabs/puppet_litmus/blob/master/lib/puppet_litmus/inventory_manipulation.rb#L15

spec/lib/puppet_litmus/serverspec_spec.rb
```diff
@@ -50,4 +50,24 @@ class DummyClass
       end
     end
   end
+
+  describe '.bolt_upload_file' do
+    let(:local) { '/tmp' }
+    let(:remote) { '/remote_tmp' }
+    let(:result) { ['status' => 'success', 'result' => { 'exit_code' => 0, 'stdout' => nil, 'stderr' => nil }] }
+    let(:inventory_hash) { Hash.new(0) }
```
ThoughtCrhyme: There is no test for if inventory_hash is nil, which it will be when, as above, TARGET_HOST is set to 'localhost'.

lib/puppet_litmus/serverspec.rb
tphoney: hadnt thought about running bolt against localhost

PR #137, opened 2019-06-18
(FM-8178) Make params passable using provision.yaml

michaeltlombardi:
Prior to this commit the only way to pass additional provisioning params
to a provisioning task was via environment variables. This commit adds
an optional params hash to the provision.yaml definition, enabling a
user to pass an arbitrary set of keys to the provisioner.

These are transformed into environment variables as `LITMUS_param` where
the `param` is upcased. This will enable users of the vagrant provisioner
on windows to pass SMB username and passwords via the provision file
instead of typing them into the terminal.

The environment variables created during the run do not persist outside
of the ruby runspace.


michaeltlombardi:
Sorry, shot this off right before logging off, didn't check rubocop - fixed now though, I think!

PR #139, opened 2019-06-20
(FM-8273) Enable parallel acceptance from Windows

michaeltlombardi:
Prior to this commit running `litmus:acceptance:parallel` from a windows
node failed during the call to run the tests per target host. This was
due to the command string prepending the target host as an environment
variable before the execution, which does not work on Windows.

This commit updates the payloads creation to standardize the exec call
and enable passing an options hash, including environment variables to
send to the Open3 call.

This enables execution from a Windows node.

Co-authored-by: Glenn Sarti <glenn.sarti@outlook.com>

### DIFF
--- a/lib/puppet_litmus/rake_tasks.rb
+++ b/lib/puppet_litmus/rake_tasks.rb
@@ -358,22 +358,30 @@ namespace :litmus do
         end
         spinners = TTY::Spinner::Multi.new("Running against #{targets.size} targets.[:spinner]", frames: ['.'], interval: 0.1)
         payloads = []
+        # Generate list of targets to provision
         targets.each do |target|
-          test = "TARGET_HOST=#{target} bundle exec rspec ./spec/acceptance --format progress"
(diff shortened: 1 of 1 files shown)


lib/puppet_litmus/rake_tasks.rb
```diff
@@ -358,22 +358,30 @@ def run_local_command(command)
         end
         spinners = TTY::Spinner::Multi.new("Running against #{targets.size} targets.[:spinner]", frames: ['.'], interval: 0.1)
         payloads = []
+        # Generate list of targets to provision
         targets.each do |target|
-          test = "TARGET_HOST=#{target} bundle exec rspec ./spec/acceptance --format progress"
+          test = 'bundle exec rspbundle exec rspec ./spec/acceptance --format progress'
```
tphoney: `rspbundle` looks like a typo

michaeltlombardi: Whoops! fixed now.

PR #143, opened 2019-06-24
(maint) Added support to install_module for a puppetfile

gregohardy:
and added spec tests for install_module

### DIFF
--- a/lib/puppet_litmus/rake_tasks.rb
+++ b/lib/puppet_litmus/rake_tasks.rb
@@ -7,6 +7,7 @@ require 'open3'
 require 'pdk'
 require 'json'
 require 'parallel'
+require 'pry'
 
 # helper methods for the litmus rake tasks
 module LitmusRakeHelper
@@ -253,7 +254,7 @@ namespace :litmus do
   #
   # @param :target_node_name [Array] nodes on which to install a puppet module for testing.
   desc 'install_module - build and install module'
-  task :install_module, [:target_node_name] do |_task, args|
+  task :install_module, [:target_node_name, :puppetfile] do |_task, args|
     inventory_hash = inventory_hash_from_inventory_file
     target_nodes = find_targets(inventory_hash, args[:target_node_name])
     if target_nodes.empty?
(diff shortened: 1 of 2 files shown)


lib/puppet_litmus/rake_tasks.rb
```diff
@@ -253,46 +254,74 @@ def run_local_command(command)
...
+
+    if args[:puppetfile]
+      run_local_command('mkdir -p ~/.puppetlabs/bolt/')
+      run_local_command('mkdir -p ~/.puppetlabs/bolt/modules')
+      run_local_command("cp #{args[:puppetfile]} ~/.puppetlabs/bolt/Puppetfile")
+      run_local_command('bolt puppetfile install')
+
+      # puppet module install defaults to the first path in the module path. Check the modulepath on the remote
+      # with a call. We will use that to install our modules.
+      output = run_command('puppet module list', target_nodes, config: nil, inventory: inventory_hash)
```
tphoney: Is it possible to set the module path ? because if you are running tests for multiple modules it will install loads of modules. ie can the module install directory be local to the working directory, eg ./.module/

lib/puppet_litmus/rake_tasks.rb
```diff
@@ -253,46 +254,74 @@ def run_local_command(command)
...
+
+      # modulepath should be an optional arg
+      modulepath = ''
+      begin
+        modulepath = output[0]['result']['stdout'].split("\n")[0].split[0]
+      rescue StandardError => e
+        raise "No modulepath detected on the remote host [#{target_nodes}] error [#{e}]"
       end
+
+      run_local_command("bundle exec bolt file upload ~/.puppetlabs/bolt/modules #{modulepath} --nodes #{target_string} --inventoryfile inventory.yaml")
```
tphoney: windows is not going to like this at all. see my above comment.

lib/puppet_litmus/rake_tasks.rb
```diff
@@ -7,6 +7,7 @@
 require 'pdk'
 require 'json'
 require 'parallel'
+require 'pry'
```
tphoney: dont need this

lib/puppet_litmus/rake_tasks.rb
```diff
@@ -253,46 +254,74 @@ def run_local_command(command)
...
+      # raise "Failed to run 'pdk_build_command',#{stdout} and #{stderr}" if (stderr =~ %r{completed successfully}).nil?
+      require 'pdk/module/build'
+      opts = {}
+      opts[:force] = true
+      builder = PDK::Module::Build.new(opts)
+      module_tar = builder.build
+      raise "Unable to find package in 'pkg/*.tar.gz'" if module_tar.nil?
+
+      puts 'Built'
+      run_local_command("bundle exec bolt file upload #{module_tar} /tmp/#{File.basename(module_tar)} --nodes #{target_string} --inventoryfile inventory.yaml")
```
tphoney: https://github.com/puppetlabs/puppet_litmus/blob/54ae717c6f1f1a293df55b16d128f5add807255b/lib/puppet_litmus/serverspec.rb#L120 rather than run a command we can call boltspec directly https://github.com/puppetlabs/puppet_litmus/blob/54ae717c6f1f1a293df55b16d128f5add807255b/lib/puppet_litmus/serverspec.rb#L120

lib/puppet_litmus/rake_tasks.rb
tphoney: https://github.com/puppetlabs/puppet_litmus/blob/54ae717c6f1f1a293df55b16d128f5add807255b/lib/puppet_litmus/serverspec.rb#L120 instead of shelling out

lib/puppet_litmus/rake_tasks.rb
```diff
@@ -253,46 +254,74 @@ def run_local_command(command)
...
+
+      install_module_command = "puppet module install /tmp/#{File.basename(module_tar)}"
+      result = run_command(install_module_command, target_nodes, config: nil, inventory: inventory_hash)
+
+      # rubocop:disable Style/GuardClause
```
tphoney: we should take out the rubocop disable ` raise "Failed trying to run '#{install_module_command}' against inventory." 
 unless result.is_a?(Array)
  result.each do |node|
          puts "#{node['node']} failed #{node['result']}" if node['status'] != 'success'
        end`

lib/puppet_litmus/rake_tasks.rb
gregohardy: Yup. It's one more thing to set. I used that method as it is what puppet module install does. If you don't specify it just resorts to using the first path in the module path.

lib/puppet_litmus/rake_tasks.rb
gregohardy: That's in the original code. You want it removed still?

lib/puppet_litmus/rake_tasks.rb
gregohardy: Cool.

PR #145, opened 2019-06-25
(FM-8284) Add bolt_run_script command

eimlav:
Adds a command which can be used to run a script file against a target
host.

### DIFF
--- a/lib/puppet_litmus/serverspec.rb
+++ b/lib/puppet_litmus/serverspec.rb
@@ -158,4 +158,30 @@ module PuppetLitmus::Serverspec
     yield result if block_given?
     result
   end
+
+  # Runs a script against the target system.
+  #
+  # @param script [String] The path to the script on the source machine
+  # @param opts [Hash] Alters the behaviour of the command. Valid options are :expect_failures [Boolean] doesnt return an exit code of non-zero if the command failed.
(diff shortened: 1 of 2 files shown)


lib/puppet_litmus/serverspec.rb
```diff
@@ -158,4 +158,29 @@ def run_bolt_task(task_name, params = {})
...
+  # @return [Object] A result object from the script run.
+  def bolt_run_script(script, opts = {})
+    target_node_name = ENV['TARGET_HOST'] if target_node_name.nil?
+    inventory_hash = if target_node_name.nil? || target_node_name == 'localhost'
+                       nil
+                     else
+                       inventory_hash_from_inventory_file
+                     end
+
+    result = run_script(script, target_node_name, options: opts, config: nil, inventory: inventory_hash)
```
florindragos: There's also a parameter for arguments: https://github.com/puppetlabs/bolt/blob/master/lib/bolt_spec/run.rb#L66

eimlav: @florindragos Good spot. I'll add in an arguments parameter.

eimlav: @florindragos Now updated with arguments.

PR #146, opened 2019-06-25
FM-8286 run_bolt_task returns bolt result object

sheenaajay:
Request for a review
Following changes are made to the function run_bolt_task

- Added unit testcase to verify the success and failure conditions
- Added version and _error to verify the task output for the package module
- Added expect_failure options

### DIFF
--- a/lib/puppet_litmus/serverspec.rb
+++ b/lib/puppet_litmus/serverspec.rb
@@ -137,15 +137,16 @@ module PuppetLitmus::Serverspec
   #
   # @param task_name [String] The name of the task to run.
   # @param params [Hash] key : value pairs to be passed to the task.
-  # @return [Object] A result object from the task.
-  def run_bolt_task(task_name, params = {})
+  # @param opts [Hash] Alters the behaviour of the command. Valid options are :expect_failures [Boolean] doesnt return an exit code of non-zero if the command failed.
+  # @return [Object] A result object from the task.The values available are stdout, stderr and result.
+  def run_bolt_task(task_name, params = {}, opts = {})
     config_data = { 'modulepath' => File.join(Dir.pwd, 'spec', 'fixtures', 'modules') }
     inventory_hash = inventory_hash_from_inventory_file
     target_node_name = ENV['TARGET_HOST'] if target_node_name.nil?
 
     result = run_task(task_name, target_node_name, params, config: config_data, inventory: inventory_hash)
 
-    raise "task failed\n`#{task_name}`\n======\n#{result}" if result.first['status'] != 'success'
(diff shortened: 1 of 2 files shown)


lib/puppet_litmus/serverspec.rb
```diff
@@ -138,14 +138,14 @@ def bolt_upload_file(source, destination, options = {})
   # @param task_name [String] The name of the task to run.
   # @param params [Hash] key : value pairs to be passed to the task.
```
tphoney: you will need to add in docs for the new parameter.

lib/puppet_litmus/serverspec.rb
```diff
@@ -138,14 +138,14 @@ def bolt_upload_file(source, destination, options = {})
...
   # @return [Object] A result object from the task.
```
tphoney: and add in what new values are available in the result object.

lib/puppet_litmus/serverspec.rb
```diff
@@ -154,7 +154,9 @@ def run_bolt_task(task_name, params = {})
                 end
     result = OpenStruct.new(exit_code: exit_code,
                             stdout: result.first['result']['status'],
-                            stderr: result.first['result']['status'])
+                            stderr: result.first['result']['stderr'],
+                            version: result.first['result']['version'],
+                            _error: result.first['result']['_error'])
```
tphoney: rather than return a couple of values, could we return the whole bolt result hash, as a thing ?

lib/puppet_litmus/serverspec.rb
sheenaajay: @tphoney updated the docs.

lib/puppet_litmus/serverspec.rb
sheenaajay: @tphoney updated the result section. The values available are stdout, stderr and result

lib/puppet_litmus/serverspec.rb
sheenaajay: @tphoney thanks for the valuable comment.Sounds good.We can fetch any value from result hash.

sheenaajay:
@tphoney All comments are incorporated. Please review.

lib/puppet_litmus/serverspec.rb
```diff
@@ -154,7 +155,8 @@ def run_bolt_task(task_name, params = {})
                 end
     result = OpenStruct.new(exit_code: exit_code,
                             stdout: result.first['result']['status'],
-                            stderr: result.first['result']['status'])
+                            stderr: result.first['result']['stderr'],
+                            result: result.first['result'])
```
tphoney: nice !

PR #147, opened 2019-06-27
(FM-8296) Ensure serverspec helpers emit correctly

michaeltlombardi:
Prior to this commit the `run_bolt_task` and `bolt_upload_file` methods
did not correctly emit stdout and stderr - `run_bolt_task` also swallowed
the exit codes of a run.

This commit updates both methods to correctly capture and emit the stdout
and stderr streams and ensures that a specific exit code for a task is
captured instead of being overwritten to `255`.

The reason for these updates is that the `run_task` and `upload_file`
methods leveraged by `run_bolt_task` and `bolt_upload_file` respectively
do not return the same structured results as `run_command` and
`run script` - namely, it does not return the `exit_code`, `stderr`, and
`stdout` key-value pairs in the `result` hash.

Instead the output, exit code, and error must be extricated from the
object to be passed along.

### DIFF
--- a/lib/puppet_litmus/serverspec.rb
+++ b/lib/puppet_litmus/serverspec.rb
@@ -109,7 +109,7 @@ module PuppetLitmus::Serverspec
   # @param opts [Hash] Alters the behaviour of the command. Valid options are :expect_failures [Boolean] doesnt return an exit code of non-zero if the command failed.
   # @yieldreturn [Block] this method will yield to a block of code passed by the caller; this can be used for additional validation, etc.
   # @return [Object] A result object from the command.
-  def bolt_upload_file(source, destination, options = {})
+  def bolt_upload_file(source, destination, opts = {}, options = {})
     target_node_name = ENV['TARGET_HOST'] if target_node_name.nil?
     inventory_hash = if target_node_name.nil? || target_node_name == 'localhost'
                        nil
@@ -119,16 +119,23 @@ module PuppetLitmus::Serverspec
 
     result = upload_file(source, destination, target_node_name, options: options, config: nil, inventory: inventory_hash)
 
-    raise "upload file failed\n======\n#{result}" if result.first['status'] != 'success'
+    result_obj = {
+      exit_code: 0,
+      stdout: result.first['result']['_output'],
+      stderr: nil,
+      result: result.first['result'],
+    }
 
-    exit_code = if result.first['status'] == 'success'
-                  0
-                else
-                  255
-                end
-    result = OpenStruct.new(exit_code: exit_code,
-                            stdout: result.first['result']['status'],
-                            stderr: result.first['result']['status'])
+    if result.first['status'] != 'success'
+      raise "upload file failed\n======\n#{result}" if opts[:expect_failures] != true
+
+      result_obj[:exit_code] = 255
+      result_obj[:stderr]    = result.first['result']['_error']['msg']
+    end
+
+    result = OpenStruct.new(exit_code: result_obj[:exit_code],
+                            stdout: result_obj[:stdout],
+                            stderr: result_obj[:stderr])
     yield result if block_given?
     result
   end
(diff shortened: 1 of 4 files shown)


lib/puppet_litmus/serverspec.rb
```diff
@@ -146,17 +152,23 @@ def run_bolt_task(task_name, params = {}, opts = {})
...
+    result_obj = {
+      exit_code: 0,
+      stdout: (result.first['result']['_output'] unless result.first['result']['_output'].nil?),
+      stderr: nil,
+      result: result.first['result'],
+    }
+
+    if result.first['status'] != 'success'
+      raise "task failed\n`#{task_name}`\n======\n#{result}" if opts[:expect_failures] != true
+      result_obj[:exit_code] = result.first['result']['_error']['details']['exitcode'].nil? ? 255 : result.first['result']['_error']['details']['exitcode']
```
michaeltlombardi: I don't love this ternary, could smush into another if?

lib/puppet_litmus/serverspec.rb
```diff
@@ -146,17 +152,23 @@ def run_bolt_task(task_name, params = {}, opts = {})
 
     result = run_task(task_name, target_node_name, params, config: config_data, inventory: inventory_hash)
 
-    raise "task failed\n`#{task_name}`\n======\n#{result}" if result.first['status'] != 'success' && opts[:expect_failures] != true
+    result_obj = {
+      exit_code: 0,
+      stdout: (result.first['result']['_output'] unless result.first['result']['_output'].nil?),
```
michaeltlombardi: Wasn't sure if this was idiomatic or not, but it was short and didn't require another line of code

lib/puppet_litmus/serverspec.rb
```diff
@@ -119,13 +119,19 @@ def bolt_upload_file(source, destination, options = {})
 
     result = upload_file(source, destination, target_node_name, options: options, config: nil, inventory: inventory_hash)
 
-    raise "upload file failed\n======\n#{result}" if result.first['status'] != 'success'
+    result_obj = {
+      exit_code: 0,
+      stdout: (result.first['result']['_output'] unless result.first['result']['_output'].nil?),
```
michaeltlombardi: Wasn't sure if this was idiomatic or not, but it was short and didn't require another line of code

lib/puppet_litmus/serverspec.rb
```diff
@@ -146,17 +152,23 @@ def run_bolt_task(task_name, params = {}, opts = {})
...
-    exit_code = if result.first['status'] == 'success'
-                  0
-                else
-                  255
-                end
-    result = OpenStruct.new(exit_code: exit_code,
-                            stdout: result.first['result']['status'],
-                            stderr: result.first['result']['stderr'],
-                            result: result.first['result'])
+    result = OpenStruct.new(exit_code: result_obj[:exit_code],
```
michaeltlombardi: Not super important but I was wondering why we do this instead of just returning a hash?

tphoney: because beaker :(, any tests written rely on a result object. and it would have meant large scale changes across module tests.

lib/puppet_litmus/serverspec.rb
tphoney: i never really liked ternery's, but that is a personal thing. this is 👍

ekohl: Would this work:

```suggestion
      result_obj[:exit_code] = result.first['result']['_error']['details'].fetch('exitcode', 255)
```

lib/puppet_litmus/serverspec.rb
ekohl: What does this return if it is nil? Isn't it just nil anyway?

michaeltlombardi: You're correct, good catch!

lib/puppet_litmus/serverspec.rb
michaeltlombardi: The fetch will error if details doesn't exist (and thus, the exit code is nil). 🤔 

Might be worth looking into later, but gonna leave it as-is for now, I think?

tphoney:
@ekohl are you happy with this ? also thanks for the review.

lib/puppet_litmus/serverspec.rb
ekohl: I think that will also be an issue with a hash:
```
[1] pry(main)> x = {}
=> {}
[2] pry(main)> x['result']['_error']['details']
NoMethodError: undefined method `[]' for nil:NilClass
from (pry):2:in `__pry__'
```

michaeltlombardi: @ekohl so digging into it a bit, that error comes whenever a prior key in the chain is missing - in your case, because there's no `result`, the error comes up from trying to index into nil for `_error`.

So! You're right, the code you submitted is exactly as guarded and shorter, so gonna flip it!

Thanks!

Issue #144, opened 2019-06-25
Fix spinners on non-ansi terminals (windows, CI)


glennsarti:
The spinners in Windows terminals are particularly annoying and mess up output e.g.

```
C:\Source\puppet-strings [use-litmus +2 ~2 -0 !]> bundle exec rake "litmus:provision[docker, centos:7]"
C:/Source/puppet-strings/.bundle/windows/ruby/2.5.0/gems/puppet-6.5.0-x64-mingw32/lib/puppet/util/windows/api_types.rb:6: warning: already initialized constant FFI::WIN32_FALSE
C:/Source/puppet-strings/.bundle/windows/ruby/2.5.0/gems/pdk-1.10.0/lib/pdk/util/windows/api_types.rb:6: warning: previous definition of WIN32_FALSE was here
C:/Source/puppet-strings/.bundle/windows/ruby/2.5.0/gems/puppet-6.5.0-x64-mingw32/lib/puppet/util/windows/api_types.rb:9: warning: already initialized constant FFI::ERROR_SUCCESS
C:/Source/puppet-strings/.bundle/windows/ruby/2.5.0/gems/pdk-1.10.0/lib/pdk/util/windows/api_types.rb:9: warning: previous definition of ERROR_SUCCESS was here
 Provisioning centos:7 using docker provisioner.[|]Cloning into 'spec/fixtures/modules/facts'...
Cloning into 'spec/fixtures/modules/provision'...
Cloning into 'spec/fixtures/modules/puppet_agent'...
 Provisioning centos:7 using docker provisioner.[\]remote: Enumerating objects: 34, done.
remote: Counting objects:  88% (30/34)                                                                                        remote: Counting objects:   8% (6/69)                                                                                         remote: Counting objects: 100% (69/69), done.
remote: Compressing objects: 100% (54/54), done.
remote: Compressing objects: 100% (28/28), done.
Receiving objects:  88% (30/34)    15 (delta 2), pack-reused 0 eceiving objects:  85% (29/34)
Receiving objects: 100% (34/34), 21.28 KiB | 4.25 MiB/s, done.
Resolving deltas: 100% (7/7), done.
remote: Enumerating objects: 182, done.Rremote: Total 69 (delta 7), reused 48 (delta 4), pack-reused 0                        Receiving objects: 100% (69/69), 28.49 KiB | 140.00 KiB/s, done.


Resolving deltas: 100% (7/7), done.82)    esolving deltas:   0% (0/7)
[...]

PR #149, opened 2019-06-27
(FM-8299) spinner runs after spec_prep in provision

tphoney:
the text of spec_prep got garbled by the spinner, this starts the spinner after spec_prep


tphoney:
relates to #144

florindragos:
👍

PR #151, opened 2019-07-01
(MAINT) Release prep 0.7.1

michaeltlombardi:


michaeltlombardi:
Have tested against MOTD and scheduled_task locally.

PR #160, opened 2019-07-13
(feat) Add support for append_cli parameter of provisioner

hajee:


tphoney:
I think there is an opportunity for a more generalised solution. ie allow the user to provide a params string, which we could convert into a ruby hash and passed to the bolt task. https://github.com/puppetlabs/provision/blob/master/tasks/vmpooler.rb#L10-L12 with this pr
https://github.com/puppetlabs/provision/pull/52

hajee:
I always like more generalised ;-). The snippet is from the provisioner side. What is your plan on how to add this to the rake task? Do you want the rake task to have some yaml string as a last parameter?

tphoney:
Hi @hajee i was thinking something like #162 code not tested, but the key thing is. it would allow arbitrary parameters to be passed to the provision task

PR #164, opened 2019-07-29
(feat) Add option to run acceptance in serial

michaeltlombardi:
This commit adds the `litmus:acceptance:serial` task which is a meta task
that calls the `litmus:acceptance:host_name:port` tasks one after another.

This allows for execution of tests which share dependencies, like the bolt
inventory file, and cannot run in parallel because of that.

### DIFF
--- a/lib/puppet_litmus/rake_tasks.rb
+++ b/lib/puppet_litmus/rake_tasks.rb
@@ -426,6 +426,20 @@ namespace :litmus do
         exit 1 if failure_list.any?
       end
 
+      # Run acceptance tests against all machines in the inventory file in serial.
+      desc 'Run tests in serial against all machines in the inventory file'
+      task :serial do
+        # Iterate over all of the acceptance test tasks and invoke them;
+        # We can rely on them always being in the format `litmus:acceptance:host_name:port`
+        # The host_name might be localhost or an IP or a DNS-resolvable node.
+        prefix = 'litmus:acceptance:'
+        tasks = Rake::Task.tasks.select { |task| task.name =~ %r{^#{prefix}.+:\d+$} }
+        tasks.each do |task|
+          puts "Running acceptance tests against #{task.name[prefix.length..-1]}"
+          task.invoke
+        end
+      end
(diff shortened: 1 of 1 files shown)


lib/puppet_litmus/rake_tasks.rb
```diff
@@ -426,6 +426,19 @@ def run_local_command(command)
...
 
+      # Run acceptance tests against all machines in the inventory file in serial.
+      desc 'Run tests in serial against all machines in the inventory file'
+      task :serial do
+        # Iterate over all of the acceptance test tasks and invoke them;
+        # We can rely on them always being in the format `litmus:acceptance:host_name:port`
+        # The host_name might be localhost or an IP or a DNS-resolvable node.
+        tasks = Rake::Task.tasks.select { |task| task.name =~ %r{^litmus:acceptance:.+:\d+$} }
+        tasks.each do |task|
+          puts "Running acceptance tests against #{task.name[18..-1]}"
```
ThoughtCrhyme: Why 18? What is this magic number?

michaeltlombardi: That chops off `litmus:acceptance:` from  `litmus:acceptance:host_name:port` leaving you with just `host_name:port`.

ekohl: A common way is to have a variable `prefix` and getting the length from that. It avoids these questions and errors if something changes in the prefix. If you use it in the `tasks.select` code above, it's self-documenting.

michaeltlombardi: Ah, so something like this?

```ruby
prefix = 'litmus:acceptance:'
tasks = Rake::Task.tasks.select { |task| task.name =~ %r{^#{prefix}.+:\d+$} }
tasks.each do |task|
  puts "Running acceptance tests against #{task.name[prefix.length..-1]}"
  ...
```

ekohl: Exactly

PR #166, opened 2019-08-15
MODULES-9729 update exit code for nil cases

sheenaajay:
Request for a review

### DIFF
--- a/lib/puppet_litmus/serverspec.rb
+++ b/lib/puppet_litmus/serverspec.rb
@@ -184,7 +184,11 @@ module PuppetLitmus::Serverspec
     else
       raise "task failed\n`#{task_name}`\n======\n#{result}" if opts[:expect_failures] != true
 
-      result_obj[:exit_code] = result.first['result']['_error']['details'].fetch('exitcode', 255)
+      result_obj[:exit_code] = if result.first['result']['_error']['details'].nil?
+                                 255
+                               else
+                                 result.first['result']['_error']['details'].fetch('exitcode', 255)
+                               end
       result_obj[:stderr]    = result.first['result']['_error']['msg']
     end
 


lib/puppet_litmus/serverspec.rb
```diff
@@ -184,7 +184,11 @@ def run_bolt_task(task_name, params = {}, opts = {})
     else
       raise "task failed\n`#{task_name}`\n======\n#{result}" if opts[:expect_failures] != true
 
-      result_obj[:exit_code] = result.first['result']['_error']['details'].fetch('exitcode', 255)
```
michaeltlombardi: My understanding was that this line should return 255 if nil, or the exit code if it exists. 🤔

michaeltlombardi: Do you have an example of where this has failed in an existing test?

sheenaajay: Sure @michaeltlombardi .This is the example where its failing and throws nil error. Result set doesnt have[ '_error']['details']

Fragment of testcase
params = { 'fact' => '--foo' }
result = run_bolt_task('facter_task', params, expect_failures: true)

Results
=> [{"node"=>"mhyo0hxupa8ma56.delivery.puppetlabs.net",
  "target"=>"mhyo0hxupa8ma56.delivery.puppetlabs.net",
  "action"=>"task",
  "object"=>"facter_task",
  "status"=>"failure",
  "result"=>{"_error"=>{"kind"=>"facter_task/failure", "msg"=>"Exit 1 running /opt/puppetlabs/puppet/bin/facter -p --json --foo: error: unrecognised option '--foo'\n"}}}]

michaeltlombardi: Ahhhhhhhhhh. That makes sense! I didn't know we could return an error without details. :sob:

PR #173, opened 2019-08-28
(feat) Better puppet error detection and better readable output

hajee:
The current implementation only detects failed catalog compiles. This is enough for very simple manifests, but misses errors when applying larger manifests. 

This PR uses detailed exitcodes to detect **ALL** errors. It also reports the stdout of the puppet run in a much better readable format when an error occurs.


hajee:
Is there an expected time frame when there is time to look at this PR?

florindragos:
Hi @hajee,
Thank You for your contribution!
Sorry for taking so long to review your PR. I'll have a look.

hajee:
Thanks.

PR #179, opened 2019-09-04
(FM-8342) Handle mocking of localhost

michaeltlombardi:
Prior to this commit it was not possible to mock features when running
acceptance tests against localhost. This is because both specifying the
node name localhost causes bolt to ignore the inventory file and also
litmus passed an empty inventory file when calling the helpers against
localhost as the target node.

This commit makes some extensive additions and patches to the existing
behavior in order to enable mocking localhost during testing:

- It adds localhost_inventory_hash as a method in inventory_manipulation,
  which will return a valid inventory hash containing only what bolt
  needs to execute locally; it places the node litmus_localhost in
  the local group and specifies the transport as local. This node CAN
  NOT be named localhost or bolt will ignore the inventory file.
- It adds the target_in_inventory? helper in inventory_manipulation,
  which will check to see if the specified node name is in the inventory
  and return true if it does or false if it does not.
- It adds the targeting_localhost? helper in serverspec which replaces
  the logic reused throughout the serverspec methods for determining if
  the command is being used against localhost.
- It updates the logic in the serverspec methods to set the target node
  name to litmus_localhost if the target is localhost, allowing the
  helpers to use the mockable entry in the inventory hash.
- It updates the logic in the serverspec methods to use the inventory
  file if it exists and to substitute the basic localhost inventory
  hash if it does not.
- It updates the logic in the serverspec methods to raise an error if
  the target node is not specified in the inventory hash; previously,
  specifying a target not found in the inventory hash could lead to
  errors in the actual execution which could be hard to parse.

These changes make it possible to mock features and other inventory
details when executing acceptance tests against a localhost using the
litmus library.

### DIFF
--- a/lib/puppet_litmus/inventory_manipulation.rb
+++ b/lib/puppet_litmus/inventory_manipulation.rb
@@ -18,6 +18,25 @@ module PuppetLitmus::InventoryManipulation
     inventory_hash
   end
 
+  # Provide a default hash for executing against localhost
+  #
+  # @return [Hash] inventory.yaml hash containing only an entry for localhost
+  def localhost_inventory_hash
+    {
+      'groups' => [
+        {
+          'name' => 'local',
+          'nodes' => [
+            {
+              'name' => 'litmus_localhost',
+              'config' => { 'transport' => 'local' },
+            },
+          ],
+        },
+      ],
+    }
+  end
+
   # Finds targets to perform operations on from an inventory hash.
   #
   # @param inventory_hash [Hash] hash of the inventory.yaml file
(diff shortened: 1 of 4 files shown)


michaeltlombardi:
Working my way through rubocop and spec tests, will wrap those up in the AM!

michaeltlombardi:
Okay, rubocop'd and tested.

This will enable localhost testing but will need to be paired with a small update to the `spec_helper_acceptance` file:

```ruby
if targeting_localhost?
  puts 'Running tests against this machine !'
  if File.exist?('inventory.yaml')
    inventory_hash = inventory_hash_from_inventory_file
    unless target_in_inventory?(inventory_hash, 'litmus_localhost')
      inventory_hash['groups'] = inventory_hash['groups'] | localhost_inventory_hash['groups']
      write_to_inventory_file(inventory_hash, 'inventory.yaml')
    end
  else
    write_to_inventory_file(localhost_inventory_hash, 'inventory.yaml')
  end
  if Gem.win_platform?
    set :backend, :cmd
  else
    set :backend, :exec
  end
else
  ...
```

PR #193, opened 2019-09-30
maint(fix serial task for release checks)

sheenaajay:


lib/puppet_litmus/rake_tasks.rb
```diff
@@ -377,6 +377,9 @@ def run_local_command(command)
         # The host_name might be localhost or an IP or a DNS-resolvable node.
         prefix = 'litmus:acceptance:'
         tasks = Rake::Task.tasks.select { |task| task.name =~ %r{^#{prefix}.+:\d+$} }
+        if tasks.empty?
+          tasks = Rake::Task.tasks.select { |task| task.name =~ %r{delivery.puppetlabs.net$$}; }
```
michaeltlombardi: Ahhhhhhhhhhh I understand now why we were finding them locally and failing via vmpooler! Whoops.

DavidS:
what is the format for those rake task names? how will that specific pattern interact with other backends?

lib/puppet_litmus/rake_tasks.rb
florindragos: Wouldn't it make more sense to extend the regex so it matches also for the vmpooler hostnames?

tphoney: why is it not everything but localhost ?

florindragos: something like `^#{prefix}.\S+` ?

tphoney: no tests :(

sheenaajay: @tphoney oops adding the tests now

sheenaajay: @florindragos sure let me try that. Thank you.

florindragos: actually no, this would include `serial` and `parallel`

sheenaajay:
@DavidS sorry for the late reply.
These are the formats of rake tasks
Run against vmpooler
rake litmus:acceptance:oag4qztd04qnagy.delivery.puppetlabs.net
Run against docker
rake litmus:acceptance:localhost:2222
@tphoney fixed the tests to run parallelly so we dont have to run them serially now.
Thanks a lot for looking into the PR @michaeltlombardi @florindragos @tphoney @DavidS

PR #194, opened 2019-10-01
(feature) run_bolt_task allows a custom inventory file

tphoney:


lib/puppet_litmus/serverspec.rb
```diff
@@ -24,10 +24,10 @@ def idempotent_apply(manifest)
   #
   # @param manifest [String] puppet manifest code to be applied.
   # @param opts [Hash] Alters the behaviour of the command. Valid options are:  
-  #  :catch_changes [Boolean] (false) We're after idempotency so allow exit code 0 only.
-  #  :expect_changes [Boolean] (false) We're after changes specifically so allow exit code 2 only.
-  #  :catch_failures [Boolean] (false) We're after only complete success so allow exit codes 0 and 2 only.
-  #  :expect_failures [Boolean] (false) We're after failures specifically so allow exit codes 1, 4, and 6 only.
+  #  :catch_changes [Boolean] (false) We're after idempotency so allow exit code 0 only.
```
michaeltlombardi: Trailing whitespaces

lib/puppet_litmus/serverspec.rb
```diff
@@ -171,16 +171,21 @@ def bolt_upload_file(source, destination, opts = {}, options = {})
...
+  # @param opts [Hash] Alters the behaviour of the command. Valid options are  
+  #  :expect_failures [Boolean] doesnt return an exit code of non-zero if the command failed.  
+  #  :inventory_file [String] path to the inventory file to use with the task.
   # @return [Object] A result object from the task.The values available are stdout, stderr and result.
+  # rubocop:enable Layout/TrailingWhitespace
   def run_bolt_task(task_name, params = {}, opts = {})
     config_data = { 'modulepath' => File.join(Dir.pwd, 'spec', 'fixtures', 'modules') }
     target_node_name = targeting_localhost? ? 'litmus_localhost' : ENV['TARGET_HOST']
-    inventory_hash = File.exist?('inventory.yaml') ? inventory_hash_from_inventory_file : localhost_inventory_hash
+    inventory_hash = File.exist?('inventory.yaml') ? inventory_hash_from_inventory_file(opts[:inventory_file]) : localhost_inventory_hash
```
michaeltlombardi: Shouldn't the exist check be against `opts[:inventory_file]`? Otherwise, won't it still share as long as `inventory.yaml` exists (which it likely will if folks have used litmus before in their project)?

michaeltlombardi: We do _if_ we want to be able to mock features for localhost testing, as far as I can tell.

tphoney: Uggh, more is needed. 
the logic now is good. 
we also want to add the ability to add features to the inventory file at provision time. :) this will mitigate a lot of issues

PR #196, opened 2019-10-03
(FM-8464) Remove bolt version pin

florindragos:


puppet_litmus.gemspec
```diff
@@ -20,7 +20,8 @@ Gem::Specification.new do |spec|
     Providing a simple command line tool for puppet content creators, to enable simple and complex test deployments.
   EOF
   spec.summary = 'Providing a simple command line tool for puppet content creators, to enable simple and complex test deployments.'
-  spec.add_runtime_dependency 'bolt',        ['>= 1.13.1', '< 1.29.1']
+  spec.add_runtime_dependency 'bolt',        ['>= 1.13.1', '< 2.0.0']
   spec.add_runtime_dependency 'pdk',         ['>= 1.10.0', '< 2.0.0']
   spec.add_runtime_dependency 'tty-spinner', ['>= 0.5.0', '< 1.0.0']
+  spec.add_runtime_dependency 'docker-api',  ['>= 1.34', '< 2.0.0']
```
tphoney: Is this not brought in by bolt, do we need to explicitly call this out. Do we need to bound the version ?

florindragos: Unfortunately, it is needed by the serverspec docker backend...

florindragos: It is not included in bolt anymore

PR #200, opened 2019-10-03
(FM-8316) Rename serverspec to puppet helpers

florindragos:


florindragos:
> Love it. Can we get some screen grabs of this working with real modules. to make sure nothing is broken

Reused somt old PRs to test against mysql: https://github.com/puppetlabs/puppetlabs-mysql/pull/1240
motd: https://github.com/puppetlabs/puppetlabs-motd/pull/244
(unrelated) motd unit tests are failing because of a rspec bug

PR #201, opened 2019-10-09
(bugfix) pin pdk gem and require tty

tphoney:


florindragos:
👍

florindragos:
tested against motd and mysql

PR #203, opened 2019-10-11
(MODULES-9998) Require pdk/util and remove pdk pin

florindragos:


lib/puppet_litmus/rake_tasks.rb
```diff
@@ -8,6 +8,7 @@
 require 'json'
 require 'parallel'
 require 'tty-spinner'
+require 'pdk/util'
```
cmccrisken-puppet: I see this is lazy loaded in this PR: https://github.com/puppetlabs/puppetlabs_spec_helper/pull/295/files

Should we move this load down?

PR #206, opened 2019-10-17
(MAINT) Ensure acceptance:localhost task also runs spec_prep

RandomNoun7:
During local testing I've found it very easy to forget to do the
spec_prep task when doing local testing. Given that there seems to be
no workflow I can think of where you would not want to do spec_prep
before calling litmus:acceptance:localhost, it seems to make sense to
just add it to the task.

### DIFF
--- a/spec/lib/puppet_litmus/rake_tasks_spec.rb
+++ b/spec/lib/puppet_litmus/rake_tasks_spec.rb
@@ -2,6 +2,8 @@
 
 require 'spec_helper'
 require 'rake'
+load File.expand_path('../../../lib/puppet_litmus/inventory_manipulation.rb', __dir__)
+include PuppetLitmus::InventoryManipulation # rubocop:disable Style/MixinUsage
 describe 'litmus rake tasks' do
   before(:all) do # rubocop:disable RSpec/BeforeAfterAll
     load File.expand_path('../../../lib/puppet_litmus/rake_tasks.rb', __dir__)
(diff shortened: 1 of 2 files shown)


RandomNoun7:
I can understand if you don't want this, but I'v been setting up a lot of test environments today and I must have stubbed my toes on this at least half a dozen times.

RandomNoun7:
Now that I think about it, this could probably be added to the other sec testing tasks like acceptance:parallel.

glennsarti:
Only downsides I can think of:

1. Is it idempotent?  (This would be a change from the "usual" practice of running the acceptance task over and over.  Now it's prep -> test -> prep -> test

2. Is it fast? Will this introduce more time into the acceptance feedback loop?

RandomNoun7:
Yes it’s idempotent. I tested it earlier when I got tired of stubbing my toe on this and made a script that did both. 

Is it fast? Depends a little bit on what you consider fast I guess, but it typically took it about 2 or 3 seconds I think to realize it didn’t need to do anything if it had already been done. 

As for the new workflow being prep -> test -> prep -> test -> prep, it’s a little more like prep -> test -> verify prep -> test -> verify prep -> test. 

For me, as I said, if there are strong objections still, that’s ok. This just seemed a decent way to remove an easy to forget stumbling block.

spec/lib/puppet_litmus/rake_tasks_spec.rb
```diff
@@ -2,6 +2,8 @@
 
 require 'spec_helper'
 require 'rake'
+load File.expand_path('../../../lib/puppet_litmus/inventory_manipulation.rb', __dir__)
+include PuppetLitmus::InventoryManipulation # rubocop:disable Style/MixinUsage
```
RandomNoun7: These two lines are in here because they are undeclared dependencies for this test file. The full suite runs because previous test file shave loaded these, but if you try to run this file on it's own these are missing.

Issue #48, opened 2018-12-20
spec_helper_acceptance example is verbose


MikaelSmith:
If we believe the example for spec/spec_helper_acceptance.rb is a good common pattern, could that be pulled into a helper that we call?

tphoney:
We have a ticket to move this into pdk-templates

ekohl:
@tphoney I think that's a bad idea. You're just copying a ton of code. pdk-templates has the same problem as modulesync: you're making it easy to have a ton of boiler plate rather than designing proper libraries. In some cases this is unavoidable (Travis config) but with Ruby code we can avoid this. [puppetlabs_spec_helper](https://github.com/puppetlabs/puppetlabs_spec_helper) is a good example of avoiding this. I'd like to either see some helpers in litmus itself or in puppetlabs_spec_helper.

tphoney:
as always, you are right @ekohl 👍 . This could just be in litmus itself, and people can override it if needs be.
My goal was to stop spec_helper_acceptance.rb to be a mess of boiler plate and customer. We could just move this as another ruby file that we load at the top of the spec_helper_acceptance.rb

DavidS:
For internal reference, the pdk-templates ticket is FM-8300 and I created https://tickets.puppetlabs.com/browse/MODULES-10031 for the move-to-litmus part.

ekohl:
https://github.com/puppetlabs/puppet_litmus/pull/210 is merged so now the docs need updating. Maybe this needs to wait for a release that includes it?

michaeltlombardi:
Docs are now updated, the docs wiki is separate from the release cadence, the latest release of Litmus includes the configure method.

https://github.com/puppetlabs/puppet_litmus/wiki/Converting-a-module-to-use-Litmus#specspec_helper_acceptancerb

PR #210, opened 2019-10-28
(FM-8355) Add spec_helper_acceptance

florindragos:
This way, we just need to `require 'litmus_spec_helper_acceptance'` in the spec_helper_acceptance of every module

### DIFF
--- a/lib/puppet_litmus.rb
+++ b/lib/puppet_litmus.rb
@@ -7,6 +7,7 @@ require 'bolt_spec/run'
 require 'puppet_litmus/inventory_manipulation'
 require 'puppet_litmus/puppet_helpers'
 require 'puppet_litmus/rake_helper'
+require 'puppet_litmus/spec_helper_acceptance'
 
 # Helper methods for testing puppet content
 module PuppetLitmus
--- a/lib/puppet_litmus/spec_helper_acceptance.rb
+++ b/lib/puppet_litmus/spec_helper_acceptance.rb
@@ -1,0 +1,90 @@
+# frozen_string_literal: true
+
+# Helper methods for testing puppet content
+module PuppetLitmus
+  extend PuppetLitmus::InventoryManipulation
+
+  def self.configure!
+    require 'serverspec'
+
+    RSpec.configure do |config|
+      config.include PuppetLitmus
+    end
+
+    if ENV['TARGET_HOST'].nil? || ENV['TARGET_HOST'] == 'localhost'
+      puts 'Running tests against this machine !'
+      if Gem.win_platform?
+        set :backend, :cmd
+      else
+        set :backend, :exec
+      end
+    else
+      # load inventory
+      inventory_hash = inventory_hash_from_inventory_file
+      node_config = config_from_node(inventory_hash, ENV['TARGET_HOST'])
+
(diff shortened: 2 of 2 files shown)


ekohl:
Related issue: https://github.com/puppetlabs/puppet_litmus/issues/48

florindragos:
> If you wrap all of this in a module function (e.g.`PuppetLitmus.configure!`) then you could write some unit tests around its behavior. Further, you could make the method take an options hash or yield a configuration object to a block to allow for more explicit customization, even if the defaults continue to look up things from the `ENV` and node config files.
> 
> Usage in `spec_helper_acceptance.rb` then looks like:
> 
> ```ruby
> require 'puppet_litmus'
> PuppetLitmus.configure!
> ```
> 
> or even:
> 
> ```ruby
> begin
>   require 'puppet_litmus'
>   PuppetLitmus.configure!
> rescue LoadError
>   puts 'Unable to load Litmus, did you add it to your Gemfile?'
> end
> ```

We would still need to do `include PuppetLitmus` to have all the helper methods available for the specs unless there is some ruby magic to include a module by calling one of its instance methods.

florindragos:
Updated code works with
```ruby
require 'puppet_litmus'

include PuppetLitmus
PuppetLitmus.configure!
```

ekohl:
Looks like there's a merge conflict so a rebase is needed.

lib/litmus_spec_helper_acceptance.rb
```diff
@@ -0,0 +1,85 @@
...
+          endpoint: endpoint,
+          operation_timeout: 300,
+        }
+
+        winrm = WinRM::Connection.new opts
+        Specinfra.configuration.winrm = winrm
+      end
+    end
+  end
+end
```
ekohl: I'm a big fan of newlines

lib/puppet_litmus.rb
```diff
@@ -7,6 +7,7 @@ module PuppetLitmus; end
 require 'puppet_litmus/inventory_manipulation'
 require 'puppet_litmus/puppet_helpers'
 require 'puppet_litmus/rake_helper'
+require 'litmus_spec_helper_acceptance'
```
ekohl: I think this should be in `puppet_litmus` as well:
```suggestion
require 'puppet_litmus/spec_helper'
```

I'm also wondering if it should be included by default.

florindragos: yeah, that would make more sense

lib/litmus_spec_helper_acceptance.rb
florindragos: so is rubocop :)

florindragos: for some reason the newline does not show up here...

ekohl: It's there. There is a special symbol Github places here if there's none.

ekohl:
With the last upgrade the `include PuppetLitmus` is no longer needed, correct?

florindragos:
> With the last upgrade the `include PuppetLitmus` is no longer needed, correct?

yep, just tested

```require 'puppet_litmus'
PuppetLitmus.configure!```

PR #214, opened 2019-11-05
(MAINT) Fix undeclared dependencies

RandomNoun7:
The gemspec does not mention that this gem requires `parallel` and
`rspec`. Most users don't notice because those gems are already brought
in by other dependencies in the typical module's Gemfile.

The issue is when someone wants to leverage Litmus, but they aren't
doing modulet testing and want to avoid installing those other module
testing gems when they aren't needed, then the Litmus gem will fail to
install if `parallel` and `rspec` are not specifically added to the
Gemfile.

This change declares those two dependencies explicitly in the gemspec to
avoid this issue.

### DIFF
--- a/puppet_litmus.gemspec
+++ b/puppet_litmus.gemspec
@@ -24,4 +24,6 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'pdk',         ['>= 1.10.0', '< 2.0.0']
   spec.add_runtime_dependency 'tty-spinner', ['>= 0.5.0', '< 1.0.0']
   spec.add_runtime_dependency 'docker-api',  ['>= 1.34', '< 2.0.0']
+  spec.add_runtime_dependency 'parallel'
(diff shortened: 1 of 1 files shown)


RandomNoun7:
I know this sounds like a bit of a weird scenario, but we're adding Litmus to a project solely for the moment because it's an easy way to get rake tasks that will provision machines either in VMPooler or with Docker.

This means that on our dev machine there actually isn't a problem because we are bringing in some of the same dev gems that other modules users are.

However, when we build the app container, we want to leave those gems out, but the rakefile still refers to litmus (require 'litmus/rake_tasks'), and so the container build blows up.

I'm working on the proper way to guard that requires statement, but in the meantime I thought explicitly calling out dependencies is usually not a bad thing.

 I don't know exactly what versions of these gems should be specified. For instance, one of the gems that usually requires parallel and avoids this issue simply specifies it's version as `>= 0`. So feel free to either push a commit to my branch that updates the version requirements, or whatever you feel is best.

PR #215, opened 2019-11-06
(maint) Lazily load gems in rake tasks

glennsarti:
Previously the Litmus rake tasks loaded every single gem whether it was needed
or not. This bloats simple commans like "rake -T". As an example, on
puppetlabs-inifile, it takes 2 seconds to just list the available rake tasks.
While this number seems initially small, when combined with all the other rake
tasks a Puppet module developer brings in, any time saving is useful.

This commit changes the loading behaviour to require gems when needed, not all
at the top of the rake_tasks.rb file. Note that some helper files were also
modified for the same lazy loading behaviour as they are consumed by the rake
tasks.


glennsarti:
In a test harness, with just an inventory.yml file.

`rake -T` went from 1.6seconds to 1.1 - Reducing runtime by 44%

PR #223, opened 2019-12-11
(FM-8779) - remove option to run acceptance tests using serial mode

lionce:
Our scope is to run acceptance tests in parallel mode to reduce the execution time.
Serial mode is not working as desired (using vmpooler provisioner is not implemented). Serial mode is not required
As a team (IA_Content team) we decided to cleanup the code and to remove it

### DIFF
--- a/lib/puppet_litmus/rake_tasks.rb
+++ b/lib/puppet_litmus/rake_tasks.rb
@@ -369,20 +369,6 @@ namespace :litmus do
         exit 1 if failure_list.any?
       end
 
-      # Run acceptance tests against all machines in the inventory file in serial.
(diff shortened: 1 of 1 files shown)


trevor-vaughan:
Does this remove the ability to run tests in serial *at all*?

If so, this puts Litmus out of the running as a Beaker replacement for my users (which is fine, I just want to make sure I know which way the wind is blowing).

DavidS:
@trevor-vaughan you can still run tests against individual nodes using the `TARGET_HOST` environment variable with `rspec`, or the `litmus:acceptance:<host>` rake tasks.

trevor-vaughan:
Ah, I see, thanks.

PR #224, opened 2019-12-11
(MODULES-10115) Fix Windows endpoint declaration

michaeltlombardi:
This commit modifies the URI definition for the serverspec endpoint to only append the default port if one is not specified. Prior to this commit the implementation appended the default port regardless leading to a broken implementation with two ports specified.

### DIFF
--- a/lib/puppet_litmus/spec_helper_acceptance.rb
+++ b/lib/puppet_litmus/spec_helper_acceptance.rb
@@ -73,7 +73,9 @@ module PuppetLitmus
         set :os, family: 'windows'
         user = node_config.dig('winrm', 'user') unless node_config.dig('winrm', 'user').nil?
         pass = node_config.dig('winrm', 'password') unless node_config.dig('winrm', 'password').nil?
-        endpoint = "http://#{ENV['TARGET_HOST']}:5985/wsman"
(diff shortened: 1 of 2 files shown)


lib/puppet_litmus/spec_helper_acceptance.rb
```diff
@@ -73,7 +73,7 @@ def self.configure!
         set :os, family: 'windows'
         user = node_config.dig('winrm', 'user') unless node_config.dig('winrm', 'user').nil?
         pass = node_config.dig('winrm', 'password') unless node_config.dig('winrm', 'password').nil?
-        endpoint = "http://#{ENV['TARGET_HOST']}:5985/wsman"
+        endpoint = (ENV['TARGET_HOST'] =~ %r{:\d+$}) ? "http://#{ENV['TARGET_HOST']}/wsman" : "http://#{ENV['TARGET_HOST']}:5985/wsman"
```
DavidS: Rather than breaking out the old regex, let https://ruby-doc.org/stdlib-2.4.1/libdoc/uri/rdoc/URI.html  do the parsing

michaeltlombardi: Swapped to using URI to parse and manage port/endpoint.

PR #225, opened 2019-12-12
(MODULES-10018) Contain BoltSpec::Run methods

michaeltlombardi:
Prior to this commit the BoltSpec::Run methods were included at the top level of the PuppetLitmus module namespace. This caused an issue when leveraging the `command()` idiom in serverspec where it attempted to resolve `run_command()` via the BoltSpec::Run method instead of the appropriate serverspec method.

This commit contains the BoltSpec::Run methods into a class and adds a shorthand reference to them for use in the rest of the PuppetLitmus namespace.  This resolves the conflicts which prevented users from leveraging idiomatic serverspec to test their modules and systems.

### DIFF
--- a/lib/puppet_litmus.rb
+++ b/lib/puppet_litmus.rb
@@ -11,7 +11,18 @@ require 'puppet_litmus/spec_helper_acceptance'
 
 # Helper methods for testing puppet content
 module PuppetLitmus
-  include BoltSpec::Run
+  # Container class for BoltSpec::Run methods to avoid leaking them via include
+  class BoltSpecRun
+    include BoltSpec::Run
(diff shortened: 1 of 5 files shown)


DavidS:
Would this change require changing existing tests that use `BoltSpec`? If yes, is there a way to avoid that and scope the change smaller, e.g. by `include`ing `PuppetLitmus` only in examples via https://relishapp.com/rspec/rspec-core/v/3-9/docs/helper-methods/define-helper-methods-in-a-module#include-a-module-in-all-example-groups

michaeltlombardi:
Deprecated in favor of #226

PR #226, opened 2019-12-12
(MODULES-10018) Extend RSpec config with PuppetLitmus

michaeltlombardi:
Prior to this commit the configuration for RSpec as defined in the spec_helper_acceptance included PuppetLitmus. This commit also extends the configuration with PuppetLitmus.

This enables test writers to use Litmus helpers such as `apply_manifest` or `run_shell` as describe subjects.

### DIFF
--- a/lib/puppet_litmus/spec_helper_acceptance.rb
+++ b/lib/puppet_litmus/spec_helper_acceptance.rb
@@ -9,6 +9,7 @@ module PuppetLitmus
 
     RSpec.configure do |config|
       config.include PuppetLitmus
+      config.extend PuppetLitmus
     end
 
     if ENV['TARGET_HOST'].nil? || ENV['TARGET_HOST'] == 'localhost'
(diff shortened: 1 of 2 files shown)


michaeltlombardi:
This only helps if the `spec_helper_acceptance.tb` of the dependent module uses the configure method, like:

```ruby
require 'puppet_litmus'
require 'spec_helper_acceptance_local' if File.file?(File.join(File.dirname(__FILE__), 'spec_helper_acceptance_local.rb'))

PuppetLitmus.configure!
```

Note that, if you need to use the module methods in your `spec_helper_acceptance_local.rb` you will need to include litmus in a singleton class like:

```ruby
require 'puppet_litmus'
require 'singleton'

class Helper
  include Singleton
  include PuppetLitmus
end

def some_helper_method
  Helper.instance.bolt_run_script('path/to/file')
end
```

PR #233, opened 2020-01-23
(feat) move to v2 of bolt inventory file

tphoney:


lib/puppet_litmus/inventory_manipulation.rb
```diff
@@ -45,13 +48,11 @@ def localhost_inventory_hash
...
   def find_targets(inventory_hash, targets)
-    require 'bolt/inventory'
-    if targets.nil?
-      inventory = Bolt::Inventory.new(inventory_hash, nil)
-      targets = inventory.node_names.to_a
-    else
-      targets = [targets]
-    end
+    targets = if targets.nil?
+                inventory_hash.to_s.scan(%r{uri"=>"(\S*)"}).flatten
```
DavidS: Need to flag this up with the bolt team to find new API to replace `Bolt::Inventory.new(inventory_hash, nil)` with a v2 inventory.

lib/puppet_litmus/rake_helper.rb
```diff
@@ -150,7 +150,8 @@ def install_agent(collection, targets, inventory_hash)
              end
     raise "puppet_agent was not found in #{DEFAULT_CONFIG_DATA['modulepath']}, please amend the .fixtures.yml file" unless File.directory?(File.join(DEFAULT_CONFIG_DATA['modulepath'], 'puppet_agent'))
 
-    run_task('puppet_agent::install', targets, params, config: DEFAULT_CONFIG_DATA, inventory: inventory_hash)
+    # using boltspec, when the runner is called it changes the inventory_hash dropping the version field. The clone works around this
+    run_task('puppet_agent::install', targets, params, config: DEFAULT_CONFIG_DATA, inventory: inventory_hash.clone)
```
DavidS: replace this with `unmarshal(marshal(inventory_hash))` because after this I don't trust bolt any more to not screw us more over in the future

DavidS:
Marked this as backwards-incompatible, because it'll need a new version of provision (see https://github.com/puppetlabs/provision/pull/93 for the changes). This should not be user-visible, but for people having custom provision tasks or local copies of the provision module will have to look out and adapt/update.

Issue #234, opened 2020-01-28
[Feature] Remove dependency on PDK


glennsarti:
## Use Case
Currently litmus pulls in the entire PDK library purely to build a module.  As this is such a simple task, the code should be copied from PDK, and once proven, available as a separate gem to be consumed by the PDK and Litmus

## Describe the Solution You Would Like
Remove dependency on PDK and vendor the code.

## Describe Alternatives You've Considered

Keep the PDK dependency - This creates a horrible situation where the PDK calls litmus which calls PDK.  Also a lot of "bloat" for a simple single task.

## Additional Context
N/A

DavidS:
This was shipped in 0.16.0

PR #235, opened 2020-01-28
(GH-234) Remove PDK dependency

glennsarti:
Block on release of the puppet-modulebuilder gem

---

Previously Litmus depended on the entire PDK gem, however it only used the
module building capability.  This commit

* Adds the puppet-modulebuilder gem into Litmus.
* Modified the rake helper to call the new Module Builder
* Modifies the gemspec to remove PDK and add the puppet-modulebuilder gem

---

TODO

- [x] Have a release of puppet-modulebuilder gem
- [x] ~~Drop the `d DROP ME` commit which is only there for testing this PR~~

### DIFF
--- a/lib/puppet_litmus/rake_helper.rb
+++ b/lib/puppet_litmus/rake_helper.rb
@@ -71,15 +71,13 @@ module PuppetLitmus::RakeHelper
       folder_handle = Dir.open(File.join(source_folder, folder))
       next if File.symlink?(folder_handle)
 
-      opts = {}
-      opts[:module_dir] = folder_handle.path
-      opts[:'target-dir'] = File.join(Dir.pwd, 'pkg')
-      opts[:force] = true
+      module_dir = folder_handle.path
+      target_dir = File.join(Dir.pwd, 'pkg')
       # remove old build folder if exists, before we build afresh
-      FileUtils.rm_rf(opts[:'target-dir']) if File.directory?(opts[:'target-dir'])
+      FileUtils.rm_rf(target_dir) if File.directory?(target_dir)
 
       # build_module
-      module_tar = build_module(opts)
+      module_tar = build_module(module_dir, target_dir)
       module_tars.push(File.new(module_tar))
     end
     module_tars
@@ -164,15 +162,20 @@ module PuppetLitmus::RakeHelper
     results
   end
 
-  def build_module(opts)
-    # old cli_way
-    # pdk_build_command = 'bundle exec pdk build  --force'
-    # stdout, stderr, _status = Open3.capture3(pdk_build_command)
-    # raise "Failed to run 'pdk_build_command',#{stdout} and #{stderr}" if (stderr =~ %r{completed successfully}).nil?
-    require 'pdk/module/build'
-    require 'pdk/util'
+  # @param opts Hash of options to build the module
+  # @param module_dir [String] The path of the module to build. If missing defaults to Dir.pwd
+  # @param target_dir [String] The path the module will be built into. The default is <source_dir>/pkg
+  # @return [String] The path to the built module
+  def build_module(module_dir = nil, target_dir = nil)
+    require 'puppet/modulebuilder'
 
-    builder = PDK::Module::Build.new(opts)
+    source_dir = module_dir || Dir.pwd
+    dest_dir = target_dir || File.join(source_dir, 'pkg')
+
+    builder = Puppet::Modulebuilder::Builder.new(source_dir, dest_dir, nil)
+    # Force the metadata to be read. Raises if metadata could not be found
+    _metadata = builder.metadata
+
     builder.build
   end
 
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ group :test do
   gem 'rubocop-rspec', '~> 1.38'
   gem 'codecov', '~> 0.1'
   gem 'simplecov', '~> 0.18'
- end
+end
 
 group :development do
   # TODO: Use gem instead of git. Section mapping is merged into master, but not yet released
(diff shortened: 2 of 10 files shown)


DavidS:
great to see this work started, it'll make the litmus install lighter and more robust.

What I'm unclear on: I thought we were talking about putting the `ModuleBuilder` in a external gem so it can be re-used by e.g. puppet-blacksmith. Not the least I'd also feel uncomfortable maintaining this code and its compatibility with the rest of the ecosystem. Did something change in that thinking?

DavidS:
PS: where does the `GH-xyz` syntax come from? it's not linking automatically to #234 and that made me miss the background description on the first go-around.

glennsarti:
Heya @DavidS 

TLDR; Prove the code and use-case in Litmus now, getting fast feedback.  In preparation to make it a gem.

---

> What I'm unclear on: I thought we were talking about putting the ModuleBuilder in a external gem so it can be re-used by e.g. puppet-blacksmith. 

We still are, but the work is a while off, at least for the PDK maintainers team. Thinking is

1. Make litmus more compatible now instead of later. Litmus is basically pulling in 15 gems (plus those gem's dependencies) for the sake of 323 LOC. Yes duplicating code can be bad, but it's not intrinsically bad. The extra gems seems like a heavy price to pay for something as (mostly) simple as tarring a directory

2. PDK currently uses beaker for package testing. Ideally we'd like to use Litmus.  We _can't_ (or shouldn't) really do that if Litmus pulls in PDK. This enables us (The PDK team) to consume Litmus

3. By getting the code into Litmus early, the PDK team can get faster feedback (aka use-in-anger) about what a Puppet Module Builder gem needs to look like. So we can release it quicker.

>  I'd also feel uncomfortable maintaining this code and its compatibility with the rest of the ecosystem.

I did consider this before I raised the PR, and it came down to balancing maintenance of somewhat duplicate code vs keeping the overhead of the PDK dependency for the sake of a one piece of functionality.  What swayed me over was:

As part of moving it out of PDK into, and eventually, it's own library is to reduce the scope of the classes/modules.  The S in SOLID. So the idea is there should be even less maintenance than if it was in the PDK itself.

The code in question hasn't had a bugfix since Jul 2019, and Oct 2018 before that.  It's relatively stable. So in the short term (say, next 6 months) the maintenance overhead I expect would be practically zero.  

Happy to have a chat about in person if you wish.

glennsarti:
> PS: where does the `GH-xyz` syntax come from? it's not linking automatically to #234 and that made me miss the background description on the first go-around.

I _think_ I originally got it from an Open Source project within Puppetlabs.  Maybe the windows modules way back when. It denoted the issue came from GitHub instead of Jira (MODULES-  PUPPET- PDK-)

I _think_ it pre-dated the `#number` syntax in github.

DavidS:
Thanks for the in-depth response. All of that makes a lot of sense!

Since litmus is still very experimental[tm] and I'm about to set up a new gem for a different thing, I'd be happy to take the code from here and release it as APL2.0 puppetlabs/puppet-module-builder 0.1.0 in the next few days.

I even expect testing this code to be easier in a stand-alone repo.

glennsarti:
Okay!  As you can see from the TODO I still need to add the unit tests for this class.  I know the acceptance tests have "proved" the code works, but I'd still like to have them.

ekohl:
I'm fully with @DavidS on this and I've wanted a standalone gem for that part for some time, precisely for puppet-blacksmith. Also overall :+1: for dropping the PDK dependency in this gem.

DavidS:
Created https://tickets.puppetlabs.com/browse/MODULES-10515 to get the infrastructure work scheduled in the team. might be a day or three until we get around to it.

glennsarti:
@DavidS Just chasing this up

DavidS:
@glennsarti @sheenaajay started on MODULES-10515 (neé IAC-532) yesterday.

tphoney:
❤️

lib/puppet_litmus/rake_helper.rb
```diff
@@ -164,15 +164,20 @@ def configure_path(inventory_hash)
...
-    # stdout, stderr, _status = Open3.capture3(pdk_build_command)
-    # raise "Failed to run 'pdk_build_command',#{stdout} and #{stderr}" if (stderr =~ %r{completed successfully}).nil?
-    require 'pdk/module/build'
-    require 'pdk/util'
+  # @param opts Hash of options to build the module
+  # @option opts [String] :module_dir The path of the module to build. If missing defaults to Dir.pwd
+  # @option opts [String] :target_dir The path the module will be built into. The default is <source_dir>/pkg
+  # @return [String] The path to the built module
+  def build_module(opts = {})
+    require 'puppet-modulebuilder'
```
florindragos: looks like this needs to be `require 'puppet/modulebuilder'`

lib/puppet_litmus/rake_helper.rb
```diff
@@ -164,15 +164,20 @@ def configure_path(inventory_hash)
...
-    # pdk_build_command = 'bundle exec pdk build  --force'
-    # stdout, stderr, _status = Open3.capture3(pdk_build_command)
-    # raise "Failed to run 'pdk_build_command',#{stdout} and #{stderr}" if (stderr =~ %r{completed successfully}).nil?
-    require 'pdk/module/build'
-    require 'pdk/util'
+  # @param opts Hash of options to build the module
+  # @option opts [String] :module_dir The path of the module to build. If missing defaults to Dir.pwd
+  # @option opts [String] :target_dir The path the module will be built into. The default is <source_dir>/pkg
+  # @return [String] The path to the built module
+  def build_module(opts = {})
```
ekohl: Any particular reason you're using an option hash instead of explicit parameters? I think explicit parameters would make the code easier to read.

florindragos: was probably taken out from the pdk: https://github.com/puppetlabs/pdk/blob/master/lib/pdk/module/build.rb#L13

PR #236, opened 2020-01-28
(feat) - new rake task to add a given feature to a group of nodes

david22swan:


david22swan:
Attempted to add spec test's but running them caused the creation of an inventory file that then caused failures.

DavidS:
add calling the task with a test feature to https://github.com/puppetlabs/puppet_litmus/blob/21fbe4ff58ddf35d70f7de53db22fde7d89e5491/.travis.yml#L39 - that'll at least run the code once

david22swan:
@DavidS Have add the call and also place a check in the method for if no feature is given

PR #238, opened 2020-01-31
(MODULES-10478) honeycomb integration to litmus

sheenaajay:
Screen Shot 2020-02-03 at 13 23 52
Request for a review
Following changes with this PR
Pointed the gem to master(waiting for the release)
Add honeycomb-beeline gem to dependencies to litmus file
Updated code to capture modulename, travis/ appveyor information
Updated code to capture infm from run_bolt_task,bolt_run_script,bolt_upload_file,run_shell(for success and failure conditions)

Appveyor is not sending traces to the honeycomb (steps are different spec_prep then litmus:acceptance:parallel )

### DIFF
--- a/lib/puppet_litmus.rb
+++ b/lib/puppet_litmus.rb
@@ -8,6 +8,7 @@ require 'puppet_litmus/inventory_manipulation'
 require 'puppet_litmus/puppet_helpers'
 require 'puppet_litmus/rake_helper'
 require 'puppet_litmus/spec_helper_acceptance'
+require 'honeycomb-beeline'
 
 # Helper methods for testing puppet content
 module PuppetLitmus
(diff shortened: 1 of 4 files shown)


lib/puppet_litmus.rb
```diff
@@ -8,11 +8,32 @@ module PuppetLitmus; end
...
+require 'honeycomb-beeline'
 
 # Helper methods for testing puppet content
 module PuppetLitmus
   include BoltSpec::Run
   include PuppetLitmus::InventoryManipulation
   include PuppetLitmus::PuppetHelpers
   include PuppetLitmus::RakeHelper
+  Honeycomb.configure do |config|
+    # config.debug = true
```
carabasdaniel: Allow this configuration to be set from an environment variable - just in case any debugging information is needed

sheenaajay: @carabasdaniel sure removing it since its checking ENV value . https://github.com/honeycombio/beeline-ruby/blob/4713124437d4038c25aeaf618291aaca064cc094/lib/honeycomb/configuration.rb#L15-L21

PR #244, opened 2020-02-17
(MAINT) Add interpolate_powershell helper method

michaeltlombardi:
Prior to this commit running PowerShell commands against target hosts during an acceptance test run required that the author both:

1. Wrap any PowerShell commands to ensure that they would be executed both  locally and remotely in PowerShell, as bolt localhost execution does not  run in PowerShell by default, and
2. Manually verify the interpolation of the code to be executed, accounting for passing from ruby to bolt to powershell/cmd and possibly beyond.

This commit adds a helper method to be available during acceptance test runs which allows an author to pass a string for the scriptblock they want to run in PowerShell; whatever that string evaluates to will be executed.

It does this first by base-64 encoding the command, enabling the use of the `EncodedCommand` flag when invoking PowerShell.exe, which it then proceeds to do.

### DIFF
--- a/lib/puppet_litmus/util.rb
+++ b/lib/puppet_litmus/util.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+# Helper methods for testing puppet content
+module PuppetLitmus::Util
+  # Ensure that a passed command is base 64 encoded and passed to PowerShell; this obviates the need to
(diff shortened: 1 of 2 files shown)


michaeltlombardi:
This PR pulls in functionality I was manually adding to windows-supportive modules; I'm not _sure_ this is the right spot for the code to live, but it seemed the most sensible spot for the time being.

PR #255, opened 2020-02-21
use ruby's notion of success instead of guessing at system semantics

DavidS:
this might be the cause for ruby segfaults not being detected correctly


DavidS:
This did not fix detecting the segfaults :-/

DavidS:
not needed.

PR #260, opened 2020-03-05
Allow users to install modules form a private forge repository

### LINKED ISSUE #259: Allow install_module rake tasks to set module_repository
## Use Case
Litmus is unable to install modules from a private forge repository (Artifactory). 

## Describe the Solution You Would Like
Allow `rake install_module` to accept a `module_repository` argument that defaults to `https://forgeapi.puppetlabs.com`

## Describe Alternatives You've Considered
Updating the global puppet.conf file. This will imply creating a separate bolt task to bootstrap the platform and it will lock the vm image to only work with one  repository.

jmangt:
Closes #259 

This change will allow the user to pass a private forge repository when running:

* rake install_module
* rake reinstall_module
* rake install_modules_form_directory
* rake provision_install

### DIFF
--- a/lib/puppet_litmus/rake_helper.rb
+++ b/lib/puppet_litmus/rake_helper.rb
@@ -248,7 +248,7 @@ module PuppetLitmus::RakeHelper
     builder.build
   end
 
-  def install_module(inventory_hash, target_node_name, module_tar)
+  def install_module(inventory_hash, target_node_name, module_tar, module_repository = 'https://forgeapi.puppetlabs.com')
     require 'bolt_spec/run'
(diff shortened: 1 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,11 @@
+# editorconfig.org
```
cmccrisken-puppet: Could we also add this to the `.gitignore`?

jmangt:
Made requested changes to PR.

DavidS:
rebased and resolved conflicts. thanks for the change @jmangt , will merge as soon as the CI agrees.

PR #268, opened 2020-03-24
(maint) make sure that errors in the acceptance tests fail; fix code to actually pass

DavidS:
As it turned out PowerShell does not report errors from subcommands as errors by default.


DavidS:
The first commit starts failing the appceyor tests correctly (there's an issue with localhost handling). The second commit then fixes the failure.

Previously appveyor would have happily ignored the issue.

PR #275, opened 2020-03-26
(IAC-187) force installing modules

DavidS:
This makes `litmus:install_module` much safer as it ensures that it actually updates the module on the target systems. Previously it would have just failed quietly.

### DIFF
--- a/lib/puppet_litmus/rake_helper.rb
+++ b/lib/puppet_litmus/rake_helper.rb
@@ -229,6 +229,10 @@ module PuppetLitmus::RakeHelper
       span.add_field('litmus.target_node_name', target_node_name)
       span.add_field('litmus.module_tar', module_tar)
 
+      # make sure the target module is not installed
+      # otherwise `puppet module install` might silently skip it
+      uninstall_module(inventory_hash.clone, target_node_name, force: true)
+
       include ::BoltSpec::Run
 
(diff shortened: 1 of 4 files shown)


lib/puppet_litmus/rake_helper.rb
```diff
@@ -258,7 +258,7 @@ def install_module(inventory_hash, target_node_name, module_tar, module_reposito
                       target_node_name
                     end
     run_local_command("bundle exec bolt file upload \"#{module_tar}\" /tmp/#{File.basename(module_tar)} --nodes #{target_string} --inventoryfile inventory.yaml")
-    install_module_command = "puppet module install --module_repository #{module_repository} /tmp/#{File.basename(module_tar)}"
+    install_module_command = "puppet module install --force --module_repository #{module_repository} /tmp/#{File.basename(module_tar)}"
```
tphoney: with --force on will it install dependencies ?

DavidS: > Installs the module regardless of dependency tree, checksum changes, or whether the module is already installed. By default, installs the module in the default modulepath, even if the module is already installed in another directory. Does not install dependencies.

FFS!

DavidS: changed it to instead uninstall first. That should solve this.

PR #280, opened 2020-04-01
Fix module functions

DavidS:
This changes all modules to use module functions, which
reduces the load on global namespaces, but requires some
adjustment to the tests.

This fixes the 'undefined method facts_from_node' error
seen in some test suites.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,9 +26,8 @@ jobs:
     before_script:
     - |
       set -e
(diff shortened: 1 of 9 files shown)


DavidS:
See validation of this change in https://github.com/puppetlabs/puppetlabs-motd/pull/299

DavidS:
Superseded by #281 , a more robust approach that works with different deployment scenarios

PR #281, opened 2020-04-02
Fixes `undefined method facts_from_node` error from 0.18.0

DavidS:
Fold in add_platform_field to where it can access facts_from_node

In some cases litmus usage would not make facts_from_node available in the namespace add_platform_field sees. Specifically differences in spec_helper_acceptance would change that.

This change adjusts unit testing to how `configure!` initialises the namespaces, fixes the unit tests and moves the previously separate HoneycombUtils into InventoryManipulation, where it has the necessary visibility.

Closes #280.


DavidS:
Validation of this fix running in https://github.com/puppetlabs/puppetlabs-motd/pull/299

Issue #249, opened 2020-02-20
install_module() uses `/tmp` as target directory for uploads, fails on windows


glennsarti:
## Describe the Bug
The rake task has a hard-coded temp path `/tmp` which does not exist on windows by default and will fail.

https://github.com/puppetlabs/puppet_litmus/blob/e54cc1b00d6927ba5cfb2207371e23329144085a/lib/puppet_litmus/rake_helper.rb#L243

Because Litmus is not fully tested on Windows, we're not seeing that in CI.

Easy steps to repro:

```
david@zion:~/git/puppetlabs-motd (master)$ pdk bundle exec rake litmus:provision[vagrant,gusztavvargadr/windows-server]
pdk (INFO): Using Ruby 2.5.8
pdk (INFO): Using Puppet 6.15.0
Provisioning gusztavvargadr/windows-server using vagrant provisioner.[✔]
127.0.0.1:55985, gusztavvargadr/windows-server
david@zion:~/git/puppetlabs-motd (master)$ pdk bundle exec rake litmus:install_module
pdk (INFO): Using Ruby 2.5.8
pdk (INFO): Using Puppet 6.15.0
Built
rake aborted!
Installation of package /home/david/git/puppetlabs-motd/pkg/puppetlabs-motd-4.1.0.tar.gz failed.
Errors: {"127.0.0.1:55985"=>"The command failed with exit code 1"}
/opt/puppetlabs/pdk/share/cache/ruby/2.5.0/gems/puppet_litmus-0.18.1/lib/puppet_litmus/rake_helper.rb:358:in `raise_bolt_errors'
/opt/puppetlabs/pdk/share/cache/ruby/2.5.0/gems/puppet_litmus-0.18.1/lib/puppet_litmus/rake_helper.rb:248:in `block in install_module'
/opt/puppetlabs/pdk/share/cache/ruby/2.5.0/gems/honeycomb-beeline-2.0.0/lib/honeycomb/client.rb:65:in `start_span'
/opt/puppetlabs/pdk/share/cache/ruby/2.5.0/gems/puppet_litmus-0.18.1/lib/puppet_litmus/rake_helper.rb:228:in `install_module'
/opt/puppetlabs/pdk/share/cache/ruby/2.5.0/gems/puppet_litmus-0.18.1/lib/puppet_litmus/rake_tasks.rb:234:in `block (2 levels) in <top (required)>'
[...]

PR #296, opened 2020-05-22
Use default working directory for all uploads to SUTs

DavidS:
Using `/tmp` as upload target meant that `install_module` and
`create_manifest_file` would not work with windows SUTs, which do
not have that directory.

This changes those methods to not use any directory, dumping files
into the default user's home directory instead.

Fixes: #249


DavidS:
There's something wonky with travis going on. The "queue" Travis check is a duplicate of the successful one. :shrug:

pmcmaw:
LGTM. Merging. 
Thanks for this @DavidS :-)

PR #297, opened 2020-05-26
(IAC-825) - Adding net-ssh 5 dependent gems

pmcmaw:
Placing net-ssh 5 dependent gems into the gemspec. This means they are added in 1 place rather than in all modules.

### DIFF
--- a/puppet_litmus.gemspec
+++ b/puppet_litmus.gemspec
@@ -28,4 +28,6 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'rspec'
   spec.add_runtime_dependency 'honeycomb-beeline'
   spec.add_runtime_dependency 'rspec_honeycomb_formatter'
+  spec.add_runtime_dependency 'ed25519', ['>= 1.2', '< 2.0']
+  spec.add_runtime_dependency 'bcrypt_pbkdf', ['>= 1.0', '< 2.0']
 end


pmcmaw:
This change has been tested by making motd point at these changes and they have passed:
https://github.com/puppetlabs/puppetlabs-motd/runs/709501022?check_suite_focus=true

pmcmaw:
These changes have been tested on motd and are passing on github actions. 
https://github.com/puppetlabs/puppetlabs-motd/runs/713400234?check_suite_focus=true

pmcmaw:
This has been tested on puppet_conf and passed: https://github.com/puppetlabs/puppetlabs-puppet_conf/actions/runs/116986742

pmcmaw:
This was also tested on concat and agent installed with no issues.

DavidS:
This is a workaround for https://tickets.puppetlabs.com/browse/BOLT-1296. See https://github.com/puppetlabs/bolt/pull/1337 for details.

scotje:
@DavidS These deps have made the latest `puppet_litmus` releases not installable when PDK is building packages for Windows. (`ed25519` gem fails to compile)

Is that expected? Do we need to pin `litmus` to 0.18.1 in PDK packaging for now?

DavidS:
I should have expected that :facepalm: 

The real solution would be to fix puppet-runtime to make ed25519 installable on windows - that would resolve this and bolt's problem.

To paper over it, we (IAC) can move the dependency from here (where it is platform-agnostic) to puppet-module-gems and exclude it from windows systems. We've already got a PMG update in the works and can add that on top. that should land early next week.

does that work for you?

DavidS:
Created IAC-896 to track this

PR #309, opened 2020-06-11
Ignore stderr of serverspec commands by setting request_pty to false

### LINKED ISSUE #129: Problem accessing host_inventory['facter'] facts.
The converting tests page documents how to access [facts](https://github.com/puppetlabs/puppet_litmus/wiki/converting-tests-from-beaker-rspec-to-litmus#checking-facts) using the serverspec helper.

In my attempts to convert one module, I ran into the following issue.
`host_inventory['facter']` was `nil`, so indexing into it caused the rspec code to fail before the actual examples were ever run.

```ruby
describe 'my_module' do
  if host_inventory['facter']['os']['release']['major'] == '18.04'
    it 'the manifest I only want to test on 18.04 works idempotently with no errors' do
      idempotent_apply("file { '/tmp/test': ensure => present }")
    end
  end
end
```

```
Failure/Error: host_inventory['facter']['os']['release']['major'] == '18.04'
NoMethodError:
  undefined method `[]' for nil:NilClass
```

After some debugging I traced the problem to specinfra [facter.rb](https://github.com/mizzy/specinfra/blob/a50fe04ec2ba6c13f2e315cee39f0d3dca328b14/lib/specinfra/host_inventory/facter.rb#L8)

`ret.stdout` started with the line `2019-06-04 15:45:19.950758 WARN  puppetlabs.facter - \e[0;33mskipping external facts for \"/opt/puppetlabs/puppet/cache/facts.d\": No such file or directory` and parsing of the YAML was then failing.

The warning *is* emitted by `facter --puppet --yaml` on stderr, but due to the way the example spec helper configures net-ssh  [(request_pty true](https://github.com/puppetlabs/puppetlabs-motd/blob/4fbb8f18235b0494d1865d47ec1254e3b9e6f7fe/spec/spec_helper_acceptance.rb#L39)), stderr gets merged with stdout and this is [unavoidable](https://github.com/net-ssh/net-ssh/issues/649#issuecomment-451736730).

Some possible solutions/workarounds:
* Before trying to access the facts, do an empty puppet apply or otherwise create the missing facts.d directory.
```ruby
idempotent_apply('') # This creates some directories needed to run facter without warnings.
describe 'my_module' do
  if host_inventory['facter']['os']['release']['major'] == '18.04'
    it 'the manifest I only want to test on 18.04 works idempotently with no errors' do
      idempotent_apply("file { '/tmp/test': ensure => present }")
    end
  end
end
```
* patch specinfra to redirect stderr to /dev/null
```ruby
ret = backend.run_command('facter --puppet --yaml 2>/dev/null')
```
Don't think this is gonna be acceptable though as specinfra supports windows too?
* litmus could provide its own helpers for accessing facts?
* Set request_pty => false (but I assume it was set to true for a valid reason)

lswith:
Fixes #129

### DIFF
--- a/lib/puppet_litmus/spec_helper_acceptance.rb
+++ b/lib/puppet_litmus/spec_helper_acceptance.rb
@@ -66,7 +66,7 @@ module PuppetLitmus
                end
         set :host,        options[:host_name] || host
         set :ssh_options, options
-        set :request_pty, true
+        set :request_pty, false
       elsif target_in_group(inventory_hash, ENV['TARGET_HOST'], 'winrm_nodes')
         require 'winrm'
 


DavidS:
Thanks for your contribution! Let's hope it doesn't have any other weird side effects.

Issue #311, opened 2020-06-18
uninitialized constant PuppetLitmus::VERSION in 0.18.3


TheMeier:
## Describe the Bug
Version 0.18.3 always fails with:

```
Failure/Error: require 'puppet_litmus'
NameError:
  uninitialized constant PuppetLitmus::VERSION
# ./vendor/bundle/ruby/2.5.0/gems/puppet_litmus-0.18.3/lib/puppet_litmus/rake_helper.rb:15:in `<top (required)>'
# ./vendor/bundle/ruby/2.5.0/gems/puppet_litmus-0.18.3/lib/puppet_litmus.rb:9:in `require'
# ./vendor/bundle/ruby/2.5.0/gems/puppet_litmus-0.18.3/lib/puppet_litmus.rb:9:in `<top (required)>'
```

sanfrancrisko:
Thanks for reporting @TheMeier - can I ask which module you encountered this in?

DavidS:
@sanfrancrisko @TheMeier can you try the patch in #312?

TheMeier:
@DavidS the patch is workuing fine for me.

TheMeier:
> Thanks for reporting @TheMeier - can I ask which module you encountered this in?

Local in-house only modules. We use gitlab with a config like this;
All our modules who have acceptance test have this problem

``yaml
....
acceptance debian puppet 6:
  variables:
    BUNDLE_GEMFILE: Gemfile.litmus
    PUPPET_GEM_VERSION: "< 6.12.0"
  stage: acceptance
  before_script:
  - bundle -v
  - rm Gemfile.lock || true
  - gem --version
  - bundle install --without system_tests --path vendor/bundle --jobs $(nproc)
  script:
  - bundle exec rake 'litmus:provision_list[debian]'
  - bundle exec bolt command run 'apt-get install wget lsb-release dirmngr -y' --inventoryfile
    inventory.yaml --targets='localhost*'
  - bundle exec rake 'litmus:install_agent[puppet6]'
  - bundle exec rake litmus:install_module
  - bundle exec rake litmus:acceptance:parallel
  after_script:
  - bundle exec rake litmus:tear_down

``

DavidS:
One thing to note, we've "recently" started providing a `configure!` method that we use in all our modules and we haven't seen this issue locally. 

See also https://github.com/puppetlabs/litmus/pull/6 for a potential improvement.

DavidS:
[...]

PR #312, opened 2020-06-18
Protect version reporting from undefined-ness

DavidS:
In some cases when setting up litmus, `PuppetLitmus::VERSION`
is not defined, even though it is required at the start of the
file.

To avoid issues like #311, add a safety check here.

Fixes: #311


DavidS:
We'll get a release out within the next couple of days: IAC-896

PR #315, opened 2020-07-09
Add run_bolt_plan helper

RandomNoun7:
Add run_bolt_plan helper

This change adds the ability for Litmus to run plans.

Plans are different from tasks in two major ways:

1. The target to run a plan against is a parameter to the plan, not to
Bolt. This means that target host is going to be a parameter to the plan
that the plan author has named whatever they like. As such the method
signature for the underlying `run_plan` helper in Boltspec has no target
param. As such
when calling a plan using this helper, it is up to the caller to know
the name of the parameter in the plan that will accept the target URI
and include the parameter and it's value in the `params` hash.

2. A plan is not a single executable that is copied to a target host
and run. As such there are no standard `exit_code`, `stdout` or `stderr`
nodes in the return we get from the Boltspec helper. We get only
`status` and `value` (A string array) if success, or `status` and
`value` (A complex object) if failure.

For backward compatibility and for ease of testing this new helper
creates a fake exit code, and it creates `stdout` and `stderr` keys in
the return object. The exit code is simply `0` if the plan succeeded and
`1` if it failed. The `stdout` key a string which is the `value` key of
the real result joined by line breaks. `Stderr` is is simply the `msg`
property of the `value` key if the result was a failure. If the result
was a success then `stderr` is nil.

Since `target_node_name` is not passed as a parameter to the underlying
Boltspec helper, it's only purpose here, and the reason it still exists
in this new function, is solely as a check to ensure that the target you
are about to pass to the plan exists in the inventory file you have
provided.

### DIFF
--- a/lib/puppet_litmus/puppet_helpers.rb
+++ b/lib/puppet_litmus/puppet_helpers.rb
@@ -339,6 +339,72 @@ module PuppetLitmus::PuppetHelpers
     end
   end
 
+  # Runs a plan against the target system.
+  #
+  # @param plan_name [String] The name of the task to run.
+  # @param params [Hash] key : value pairs to be passed to the task.
+  # @param opts [Hash] Alters the behaviour of the command. Valid options are
+  #  :expect_failures [Boolean] doesnt return an exit code of non-zero if the command failed.
+  #  :inventory_file [String] path to the inventory file to use with the task.
+  # @return [Object] A result object from the task.The values available are stdout, stderr and result.
+  def run_bolt_plan(plan_name, params = {}, opts = {})
+    Honeycomb.start_span(name: 'litmus.run_plan') do |span|
+      ENV['HTTP_X_HONEYCOMB_TRACE'] = span.to_trace_header
+      span.add_field('litmus.plan_name', plan_name)
+      span.add_field('litmus.params', params)
+      span.add_field('litmus.opts', opts)
+
+      config_data = { 'modulepath' => File.join(Dir.pwd, 'spec', 'fixtures', 'modules') }
+      target_node_name = targeting_localhost? ? 'litmus_localhost' : ENV['TARGET_HOST']
+      inventory_hash = if !opts[:inventory_file].nil? && File.exist?(opts[:inventory_file])
+                         inventory_hash_from_inventory_file(opts[:inventory_file])
+                       elsif File.exist?('inventory.yaml')
+                         inventory_hash_from_inventory_file('inventory.yaml')
+                       else
+                         localhost_inventory_hash
+                       end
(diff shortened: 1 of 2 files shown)


RandomNoun7:
Let me know if you would like this change. If you do I'll go ahead and write the tests. If not I'll figure something else out to get this done.

Running plans would be very useful for me. My current use case is I want to execute a plan in the [peadm](https://github.com/puppetlabs/puppetlabs-peadm) module that installs Puppet Enterprise. Using this module is a best practice for installing PE as it is maintained by the people at Puppet most interested in ensuring it functions properly and because it is far more flexible than anything I could write.

The inability for Litmus to run plans makes using this module impossible.

lib/puppet_litmus/puppet_helpers.rb
```diff
@@ -303,6 +303,64 @@ def run_bolt_task(task_name, params = {}, opts = {})
...
+      span.add_field('litmus.node_name', target_node_name)
+      add_platform_field(inventory_hash, target_node_name)
+
+      bolt_result = run_plan(plan_name, params, config: config_data, inventory: inventory_hash)
+      # Plans don't provide stdin and stdout streams. They only have a result.
+      # They don't provide an exit code either, just a 'success' value.
+      result_obj = {
+        status: nil,
+        result: nil,
+      }
```
tphoney, line 341:
```
result_obj = {
        exit_code: 0,
        stdout: nil,
        stderr: nil,
        result: bolt_result.first['value'],
      }
```
The reason for this is to allow backwards compatibility for beaker style tests. 
Granted we are adding something that beaker never had, but people still write tests where they check for status code, even though there is no status code. they just want to know if it failed with a .zero? check
You should be able to copy from https://github.com/puppetlabs/puppet_litmus/blob/b4f659013855d5900f2481bd2cc9fc2f56b7af2f/lib/puppet_litmus/puppet_helpers.rb#L245

RandomNoun7: Yeah I can add those. I'll just a comment in the code that they aren't real. Just placeholders as you say for compatibility.

RandomNoun7: Can I get away with just adding the exit code and not sdtout and stderr?

RandomNoun7: Don't mind me. I decided to add those in too.

RandomNoun7:
There seem to be a good number of rubocop violations that aren't caused by my PR. I don't know if you want to wait until those are resolved and tests pass before getting this in or not.

RandomNoun7:
Ping

PR #318, opened 2020-07-24
Add optional ignore_dependencies parameter to install_module function

alanfryer:
Puppet Litmus is being used on servers that do not external access to the internet, we get errors when trying to run the Puppet Litmus Task **_install_module_**.  This ultimately executes _**puppet module install**_ to install the module which then looks to install module dependencies from Puppet Forge which fails.  To get around this we have tried using the Rake Task **_install_modules_from_directory_**, which also fails as this calls the module_install helper function as above, recusviely for all modules in the specified directory giving the same error.

I have added the optional parameter ignore_dependencies to the install_module function and the rake tasks that call it.  If the ignore_dependencies parameter is set to true then the switches --ignore-dependencies --force are added to the puppet module install command executed

### DIFF
--- a/lib/puppet_litmus/rake_helper.rb
+++ b/lib/puppet_litmus/rake_helper.rb
@@ -269,8 +269,9 @@ module PuppetLitmus::RakeHelper
   # @param target_node_name [String] the name of the target where the module should be installed
   # @param module_tar [String] the filename of the module tarball to upload
   # @param module_repository [String] the URL for the forge to use for downloading modules. Defaults to the public Forge API.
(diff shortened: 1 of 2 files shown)


adrianiurca:
Hi @alanfryer ,
I updated the rake_helper.rb file to resolve some rubocop offenses.

Kind regards,
@adrianiurca

daianamezdrea:
Hi @alanfryer, this looks good to me and I also "played" a little bit with your changes. Thank you for your contribution! Cheers!

Issue #323, opened 2020-09-29
write_file Helper


RandomNoun7:
## Use Case
A test has a string variable that needs to be written to a file in an arbitrary location on the disk of a target node

## Describe the Solution You Would Like
A new puppet_helpers function called `create_file` that would take the following parameters:
String: content
String: destination
String - Optional - Linux Only: Owner
String - Optional - Linux Only: Group

## Describe Alternatives You've Considered
Currently the best way to accomplish this is to use the `create_manifest_file` helper and then use a `run_shell` helper to move the file to where I need it to be. The drawback is that this is two separate calls and I have to reimplement this double call in every repo I need it in.

Also, I can't use `run_shell` to `echo <blah> > <file>` to create a file because the escaping issues are difficult.

DavidS:
would bolt's `upload_file` (https://www.rubydoc.info/gems/bolt/2.13.0/BoltSpec/Run#upload_file-instance_method) help? That should be available to litmus users

RandomNoun7:
The drawback there is that if I'm starting from a string variable, I have to write out my own temp file locally first, before I can use that helper. I would like to be able to pass a String variable to Litmus directly.

DavidS:
understood.

trevor-vaughan:
IMO, adding a helper to wrap the file creation/deletion around bolt would be preferable to yet another(TM) file upload method.

trevor-vaughan:
Also, allowing bolt to take a String and do the right thing would also be nice.

DavidS:
Two acceptable (to me, non-exhaustive) ways to implement `create_file` are
* create a tempfile with the contents and pass that off to bolt's `upload_file`
* modify bolt to have a `upload_file_content` method that takes a `content` instead of a `source`

trevor-vaughan:
You could just let `upload_file` take `content` kind of like the regular `file` resource.

DavidS:
> You could just let `upload_file` take `content` kind of like the regular `file` resource.

that would break every single current user of `upload_file`, no?

RandomNoun7:
I think the only way to modify the Bolt function in a way that wouldn't break everyone would be to use a new options hash key that would tell Bolt to interpret the source parameter as a content string and not a path. I just don't know if the Bolt team would actually sign off on that or not.

[...]

PR #324, opened 2020-09-29
Add write_file helper

RandomNoun7:
This commit adds a helper to upload a string variable as a file to a
target host.

This prevents the current requirement to use the manifest_file helper
and then a separate shell command to move the file to desired
destination path on the target host.

Closes: #323

### DIFF
--- a/lib/puppet_litmus/puppet_helpers.rb
+++ b/lib/puppet_litmus/puppet_helpers.rb
@@ -147,6 +147,42 @@ module PuppetLitmus::PuppetHelpers
     end
   end
 
+  # Writes a string variable to a file on a target node at a specified path.
+  #
+  # @param content [String] String data to write to the file.
+  # @param destination [String] The path on the target node to write the file.
+  # @return [Bool] Success. The file was succesfully writtne on the target.
+  def write_file(content, destination)
+    Honeycomb.start_span(name: 'litmus.write_file') do |span|
+      ENV['HTTP_X_HONEYCOMB_TRACE'] = span.to_trace_header
+      span.add_field('litmus.destination', destination)
+
+      require 'tmpdir'
+      target_node_name = ENV['TARGET_HOST']
+
+      Tempfile.create('litmus') do |tmp_file|
+        tmp_file.write(content)
+        tmp_file.flush
+        if target_node_name.nil? || target_node_name == 'localhost'
+          require 'fileutils'
+          # no need to transfer
+          FileUtils.cp(tmp_file.path, destination)
+        else
+          # transfer to TARGET_HOST
+          inventory_hash = inventory_hash_from_inventory_file
+          span.add_field('litmus.node_name', target_node_name)
+          add_platform_field(inventory_hash, target_node_name)
+
+          bolt_result = upload_file(tmp_file.path, destination, target_node_name, options: {}, config: nil, inventory: inventory_hash)
+          span.add_field('litmus.bolt_result.file_upload', bolt_result)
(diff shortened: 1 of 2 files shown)


RandomNoun7:
This is a working candidate implementation. It still needs tests, but is available to gather feedback.

lib/puppet_litmus/puppet_helpers.rb
```diff
@@ -147,6 +147,40 @@ def create_manifest_file(manifest)
...
+  # @param destination [String] The path on the target node to write the file.
+  # @return [Bool] Success. The file was succesfully writtne on the target.
+  def write_file(content, destination)
+    Honeycomb.start_span(name: 'litmus.write_file') do |span|
+      ENV['HTTP_X_HONEYCOMB_TRACE'] = span.to_trace_header
+      span.add_field('litmus.destination', destination)
+
+      require 'tmpdir'
+      target_node_name = ENV['TARGET_HOST']
+      tmp_filename = File.join(Dir.tmpdir, "litmus_tmp_#{Time.now.strftime('%Y%m%d')}_#{Process.pid}_#{rand(0x100000000).to_s(36)}")
```
DavidS, line 162: This looks insecure. Please use https://ruby-doc.org/stdlib-2.7.1/libdoc/tempfile/rdoc/Tempfile.html instead.

lib/puppet_litmus/puppet_helpers.rb
```diff
@@ -147,6 +147,40 @@ def create_manifest_file(manifest)
...
+      tmp_file = File.open(tmp_filename, 'w')
+      tmp_file.write(content)
+      tmp_file.close
```
DavidS, line 165: It also looks like this is not removing the `tmp_file` after it is done. `tempfile` will address that too.

DavidS (changes requested):
* [x] use tempfile
* [x] add at least basic unit tests to spec/lib/puppet_litmus/puppet_helpers_spec.rb

lib/puppet_litmus/puppet_helpers.rb
RandomNoun7: With the way the existing `create_manifest_file` I thought that was the intended functionality. My temp directory is littered with them. I'll fix it in this new helper but I should probably go back and put up a maint to fix the existing helper.

DavidS: yes, please :heart:

lib/puppet_litmus/puppet_helpers.rb
```diff
@@ -147,6 +147,48 @@ def create_manifest_file(manifest)
...
+          inventory_hash = inventory_hash_from_inventory_file
+          span.add_field('litmus.node_name', target_node_name)
+          add_platform_field(inventory_hash, target_node_name)
+
+          bolt_result = upload_file(tmp_file.path, destination, target_node_name, options: {}, config: nil, inventory: inventory_hash)
+          span.add_field('litmus.bolt_result.file_upload', bolt_result)
+          raise bolt_result.first['value'].to_s unless bolt_result.first['status'] == 'success'
+
+          if opts.key? :owner
+            bolt_result = run_shell("chown #{opts[:owner]} #{destination}")
```
michaeltlombardi, line 181: Is `owner` and opt you can pass to a Windows box?

RandomNoun7: Nope. This PR is an artifact of a solution I needed to make file writes on a master work. ACL's are so much more complex I figure Windows support for permissions will take it's own PR.

michaeltlombardi (approved):

PR #327, opened 2020-10-28
Calculate github actions job matrix from metadata.json; fix frozen-string modification in puppet_output

DavidS:
The `matrix_from_metadata` executable supports creating a platform
matrix for the new cloud-ci project based off the `metadata.json`
in the current module.

### DIFF
--- a/exe/matrix_from_metadata
+++ b/exe/matrix_from_metadata
@@ -1,0 +1,62 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# this script creates a build matrix for github actions from the claimed supported platforms in metadata.json
+
+require 'json'
+
+IMAGE_TABLE = {
+  'RedHat-6' => 'rhel-6-v20200618',
+  'RedHat-7' => 'rhel-7-v20200618',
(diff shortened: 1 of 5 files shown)


DavidS:
Used in https://github.com/puppetlabs/puppetlabs-testing/pull/308

exe/matrix_from_metadata
```diff
@@ -0,0 +1,62 @@
...
+  platform: [],
+  collection: %w[
+    puppet5
+    puppet6
+    puppet7-nightly
+  ],
+}
+
+metadata = JSON.parse(File.read('metadata.json'))
+metadata['operatingsystem_support'].sort { |a, b| a['operatingsystem'] <=> b['operatingsystem'] }.each do |sup|
```
ekohl, line 46: Any particular reason not to use `sort_by { |o| o['operatingsystem'] }` here? Similar below.

DavidS: haven't read the stdlib docs in a long while. Thanks for catching that.

carabasdaniel (approved):

PR #332, opened 2020-11-09
Add validation check after agent install

carabasdaniel:
Validates that puppet command is usable in the environment after a successful return from the agent install task run.


lib/puppet_litmus/rake_tasks.rb
```diff
@@ -122,6 +122,21 @@
...
+        begin
+          responses = run_command('puppet --version', targets, options: {}, config: DEFAULT_CONFIG_DATA, inventory: inventory_hash.clone)
+          responses.each do |response|
+            raise "Error checking puppet version on #{response.to_json}" if response['status'] != 'success'
+          end
+        rescue StandardError => e
+          puts "ERROR:#{e}"
+          retries += 1
+          sleep 3
+          retry if retries <= 3
```
DavidS, line 136: Are 3x3 seconds already enough for this to succeed?

DavidS (approved):

PR #345, opened 2020-11-25
Check connectivity after provision

DavidS:


DavidS:
Depends on https://github.com/puppetlabs/provision/pull/141

PR #350, opened 2020-12-01
Convert inventory vars to JSON

jarretlavallee:
Prior to this commit, the inventory_vars were passed to the provisioners
as a struct. The provisioners expect `vars` as a JSON formatted string,
which would cause failures in provisioning when `vars` was populated.
This commit changes the format of vars to be a string before it is
passed into `provision`.

### DIFF
--- a/lib/puppet_litmus/rake_helper.rb
+++ b/lib/puppet_litmus/rake_helper.rb
@@ -134,8 +134,9 @@ module PuppetLitmus::RakeHelper
   end
 
   def provision_list(provision_hash, key)
+    require 'json'
     provisioner = provision_hash[key]['provisioner']
-    inventory_vars = provision_hash[key]['vars']
+    inventory_vars = provision_hash[key]['vars'].to_json unless provision_hash[key]['vars'].nil?
     # Splat the params into environment variables to pass to the provision task but only in this runspace
     provision_hash[key]['params']&.each { |k, value| ENV[k.upcase] = value.to_s }
     results = []
--- a/lib/puppet_litmus/rake_tasks.rb
+++ b/lib/puppet_litmus/rake_tasks.rb
@@ -24,13 +24,15 @@ namespace :litmus do
   task :provision_list, [:key] do |_task, args|
     raise 'Cannot find provision.yaml file' unless File.file?('./provision.yaml')
 
+    require 'json'
+
     provision_hash = YAML.load_file('./provision.yaml')
     raise "No key #{args[:key]} in ./provision.yaml, see https://puppetlabs.github.io/litmus/Litmus-core-commands.html#provisioning-via-yaml for examples" if provision_hash[args[:key]].nil?
 
     Rake::Task['spec_prep'].invoke
 
     provisioner = provision_hash[args[:key]]['provisioner']
-    inventory_vars = provision_hash[args[:key]]['vars']
+    inventory_vars = provision_hash[args[:key]]['vars'].to_json unless provision_hash[args[:key]]['vars'].nil?
     # Splat the params into environment variables to pass to the provision task but only in this runspace
     provision_hash[args[:key]]['params']&.each { |k, value| ENV[k.upcase] = value.to_s }
     results = []
--- a/spec/lib/puppet_litmus/rake_helper_spec.rb
+++ b/spec/lib/puppet_litmus/rake_helper_spec.rb
@@ -34,6 +34,16 @@ RSpec.describe PuppetLitmus::RakeHelper do
     end
   end
 
+  context 'with provision_list and vars' do
+    let(:provision_hash) { { 'default' => { 'provisioner' => 'docker', 'images' => ['waffleimage/centos7'], 'vars' => { 'foo' => 'bar' } } } }
+    let(:results) { [] }
+
+    it 'calls function' do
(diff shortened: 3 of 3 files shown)


jarretlavallee:
I hit this while trying to leverage https://github.com/puppetlabs/provision/pull/143

The following would throw a type validation error due to the expected String type in the provisioner. 

```
viewer:
  provisioner: docker
  images: ['litmusimage/centos:7']
  vars: 
    docker_run_opts: 
      - "-p 8086:8086"
      - "-p 3000:3000"
```

Resulted in the following.

```
bundle exec rake 'litmus:provision_list[viewer]'
HEAD is now at 9972009 Merge pull request #222 from tuxmea/rel/6.1.1
HEAD is now at b071cc8 Merge pull request #152 from tuxmea/rel/3.0.1
Provisioning litmusimage/centos:7 using docker provisioner.[-]rake aborted!
Bolt::PAL::PALError: Task provision::docker:
 parameter 'vars' expects a value of type Undef or String, got Struct
```

I had to wrap the `vars` as a string to get it to work without the modifications in this PR. 

```
vars: '{"docker_run_opts": ["-p 8086:8086", "-p 3000:3000"]}'
```

lib/puppet_litmus/rake_helper.rb
```diff
@@ -134,8 +134,9 @@ def provision(provisioner, platform, inventory_vars, append_cli = nil)
   end
 
   def provision_list(provision_hash, key)
+    require 'json'
     provisioner = provision_hash[key]['provisioner']
-    inventory_vars = provision_hash[key]['vars']
+    inventory_vars = provision_hash[key]['vars'].to_json unless provision_hash[key]['vars'].nil?
```
sanfrancrisko, line 139: I'm a bit concerned about changing the format of `inventory_vars`, given the potential knock on effect to existing provisioners that may be working under the assumption of the existing format - e.g. the [provision_service provisioner](https://github.com/puppetlabs/provision/blob/main/tasks/provision_service.rb#L81). 

Litmus also allows users to create their own custom provisioners too, so again, I would be concerned if we ship this and potentially cause regressions for those folks.

Would it be possible to enhance the docker provisioner to handle this instead?

sanfrancrisko (changes requested):

lib/puppet_litmus/rake_helper.rb
jarretlavallee: I was surprised that this was the case as well. This PR should probably be closed and addressed in a different way. I have a few questions to further understand the desired formatting of `vars`.

All of the built-in provisioners have `vars` as an `Optional[String[1]]` and then is parsed within the provisioner in different ways. It is quite possible that I misunderstood what the format of `vars` should be. I was assuming it would be a hash rather than a string. What is the expected formatting of `vars`?

The following expect the string.

https://github.com/puppetlabs/provision/blob/main/tasks/provision_service.json#L23-L26 
https://github.com/puppetlabs/provision/blob/main/tasks/docker.json#L23-L26
https://github.com/puppetlabs/provision/blob/main/tasks/abs.json#L23-L26

Each provisioner task parses the data in the following. Only the `provison_service` does it differently than the others.

https://github.com/puppetlabs/provision/blob/main/tasks/vmpooler.rb#L46-L49 Expects a string of Yaml
https://github.com/puppetlabs/provision/blob/main/tasks/provision_service.rb#L83 Expects a string of Json with `;` instead of `,`. 
https://github.com/puppetlabs/provision/blob/main/tasks/docker.rb#L151-L154 Expects a string of Yaml
https://github.com/puppetlabs/provision/blob/main/tasks/docker_exp.rb#L28-L31 Expects a string of Yaml
https://github.com/puppetlabs/provision/blob/main/tasks/abs.rb#L80-L83 Expects a string of Yaml

Based on the provisioners above, most of them expect a string of yaml. Currently, litmus passes through whatever is parsed from the `provision.yaml`, which could be a string, int, array, or hash. It doesn't currently send encoded yaml though so I am not sure how that would work.

jarretlavallee:
Closing this as it is not the right approach for this problem and can cause failures in existing integrations.

lib/puppet_litmus/rake_helper.rb
sanfrancrisko: I think you've highlighted an issue here, for sure. This should likely have been defined a bit more clearly but I do believe the original aim was for it to be quite free / unrestricted so each individual provision task could handle it in the manner that was most appropriate for the provision platform. 

E.g. Docker is invoking an application on the command line, with a number of potential arguments. The provision service is communicating with a cloud based service that consumes JSON.

jarretlavallee: That makes sense. I didn't see any information on the vars and spot checked a few provisioners thinking it would give the absolute usage. I had some incorrect assumptions on how it worked and opened this PR based on those. The format of it as it stands allows for flexibility in how the variable is used in the provider. It works as-is for us now. This PR was "me not reading the code fully" error :)

A definition of what `vars` is in a doc somewhere would be great.

sanfrancrisko: I had to go looking in the code myself and have a think about it to be honest, so it definitely could be a lot clearer. I've raised #351 to address that. Appreciate the feedback.

I merged https://github.com/puppetlabs/provision/pull/143 - hopefully that unblocks the rest of your work. Thanks again for that enhancement 👍

PR #355, opened 2020-12-09
Update honeycomb trace ENV var to new name

DavidS:
This is to align this with the conversation and changes in https://github.com/honeycombio/buildevents/pull/74


DavidS:
See also https://github.com/puppetlabs/rspec_honeycomb_formatter/pull/12 which needs to be released around the same time.

carabasdaniel (approved):

PR #356, opened 2020-12-10
Add dynamic agent collections to test matrix generation

hsnodgrass:
This PR adds the ability for `matrix_from_metadata` to generate the agent collections portion of the test matrix from the module's `metadata.json`. This way, modules that don't support certain Puppet versions won't have tests run against those versions.

The version comparison isn't a full semver comparison, just a major version comparison, since the collections are generalized based on major version.

The `version_requirement` key must adhere to the format `<comparator> <version> <comparator> <version>` otherwise parsing will fail and all collections will be used. The default behavior if version parsing fails or if the correct keys don't exist is to use all collections.

### DIFF
--- a/exe/matrix_from_metadata
+++ b/exe/matrix_from_metadata
@@ -33,13 +33,15 @@ DOCKER_PLATFORMS = [
   'Ubuntu-20.04',
 ].freeze
 
+COLLECTION_MAP = {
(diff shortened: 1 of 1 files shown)


DavidS:
Closing this in favour of #357 . Please always enable "Allow changes from maintainers" to avoid this.

PR #358, opened 2020-12-14
Add support for appending to created docker command

hajee:
Sometimes you need to add extra options to the docker command. For example when you need specific volumes or a specific hostname. With the append_cli option, you can. The `docker_exp` provisioner already has support for this.

### DIFF
--- a/lib/puppet_litmus/rake_tasks.rb
+++ b/lib/puppet_litmus/rake_tasks.rb
@@ -58,14 +58,15 @@ namespace :litmus do
   #
   # @param :provisioner [String] provisioner to use in provisioning given platform.
   # @param :platform [String] OS platform for container or VM to use.
+  # @param :append_cli [String] string append to the docker command. Can be used for passing volumes or hosts names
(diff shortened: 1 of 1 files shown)


jarretlavallee:
https://github.com/puppetlabs/provision/pull/143 in the provision module may solve this problem. The PR provides the capability to add runtime arguments to docker. I use it to map some ports, but it could be used for any docker runtime arguments. 

An example of using this in the `provision.yaml` can be found at https://github.com/puppetlabs/puppet_metrics_dashboard/blob/master/provision.yaml#L17-L20

hajee:
@jarretlavallee Thanks for pointing this out. I tried if I can use this. Unfortunately not. I need the SUT to be recognized by bolt as a docker host. The docker provisioned doesn't do that.

The docker_exp provisioned unfortunately doesn't support the docker_run_opts as vars.

hajee:
It looks like during the different changes two ways have been introduced to two different providers to pass extra commands to the docker cli. I think the one in puppetlabs/provision#143 is the most versatile. I propose to close this PR and I'll make a new one that uses the same API as puppetlabs/provision#143 and removes the append_cli stuff from both litmus as the provisioners

PR #366, opened 2021-01-07
(IAC-1287) - Only log errors in debug mode

pmcmaw:
End users are only used to seeing output when using litmus if something
goes wrong. Setting so that the additional information can only be seen
when DEBUG=true is set.


DavidS (approved):

kenyon (commented):
Am I missing something, or is this PR just changing from double quotes to single quotes, which does nothing?
```
[1] pry(main)> 'true' == "true"
=> true
```

DavidS:
@kenyon thanks for spotting that! The PR here had the wrong base-branch set and didn't actually merge into main, hence the broken diff. I've posted #367 with the rebased/fixup'd change.

PR #373, opened 2021-01-20
Use puppet 6 nightlies and increase timeout limit

carabasdaniel:


DavidS (approved):

carabasdaniel:
PR with test results here: https://github.com/puppetlabs/puppetlabs-testing/runs/1734154323?check_suite_focus=true

PR #383, opened 2021-02-08
Ensure that first-order failures in idempotent_apply get reported

DavidS:
Without this change errors in the first `apply_manifest()` would not report
errors from `puppet apply`.


david22swan (approved):
It read correctly from a grammatical point of view I guess

michaeltlombardi (approved):
ope

Alvin the chipmunk wincing and smiling sympathetically

PR #385, opened 2021-02-10
Add docker images to matrix from metadata

carabasdaniel:


DavidS (changes requested):
small changes here and there; also move all changes to `_v2` as discussed on the call to avoid flagday issues

sheenaajay (approved):

DavidS (approved):

PR #394, opened 2021-03-05
use run_shell to grab os.family fact

adrianiurca:
This error appears on websphere_applicatiomn_server when integration testing is running on gcp
```shell
Failure/Error: Helper.instance.apply_manifest(pp)

NameError:
  undefined local variable or method `os' for #<Helper:0x00007fb0f819acc0>
```


DavidS (changes requested):
This error happens when the calling context is not a rspec example or context that has `os` set up. this is not a supported usecase for litmus and needs to be fixed at the call-site, not by adding an additional remote shell call to every apply_manifest

PR #403, opened 2021-04-12
(bugfix) Update inventory path

pmcmaw:
The new inventory path(`spec/fixtures/litmus_inventory.yaml`) should be used by all related functions, such as `bolt_upload_file`, `run_shell` and `apply_manifest`


pmcmaw:
This issue is causing failures in all of our modules that use integration tests.

sanfrancrisko (approved):

PR #414, opened 2021-05-24
(maint) - Increase the connection timeout limit

david22swan:
Due to recent changes the setup time for windows machines on GCP has increased and so the timeout limit needs to be raised accordingly.


pmcmaw:
LGTM.

PR #425, opened 2021-08-09
Added options to idempotent_apply

ZloeSabo:
Added a possibility to pass options to `idempotent_apply`. Without this change, developers can't use `idempotent_apply` if they want to specify a custom hiera config or a debug option.


sanfrancrisko (approved):
Thanks for the improvement @ZloeSabo

PR #427, opened 2021-08-31
Allow Litmus Functions to accept a target

### LINKED ISSUE #418: Accept one more option for opts to accept TARGET_HOST to execute against for all litmus functions
## Use Case
Accept one more option for opts to accept TARGET_HOST to execute against for all litmus functions.

So user doesn't have to add helper function in spec_helper to change target host to execute in specific nodes.
https://github.com/puppetlabs/puppetlabs-websphere_application_server/blob/main/spec/spec_helper_acceptance_local.rb#L44

## Describe the Solution You Would Like
Update the opts Hash to accept node to execute the command.
By default it will be nil. 
If value is passed then ENV['TARGET_HOST'] will be set with that node value and execute the command against it.
Old Behaviour will still be the default one. 
User will be given an option to specify the node on which they would like to execute the command.

https://github.com/puppetlabs/puppet_litmus/blob/69812a53cd44aa6e89d11cd8af367c5f56ac4dec/lib/puppet_litmus/puppet_helpers.rb#L40

Changes required for the following functions
apply_manifest
run_shell
bolt_upload_file
run_bolt_task
bolt_run_script

RandomNoun7:
(ISSUE 418) Accept a Target in Litmus Functions

This change will allow users to specify a target for each invocation of
a Litmus Puppet Helper. Specifying a target will not affect the current
value of the `ENV['TARGET_HOST']` variable.

Targets can be specified by using strings or symbols. Targets can be:
1. The name of a role assigned to a target via the `vars => roles` hash
of the target.
2. The name of a group in the inventory
3. The name of a specific target URI

The roles key in the inventory can be
1. Called either `Role` or `Roles`
2. Can have a single role `role: webserver` or an array:
  ```
  roles:
    - IIS
    - Webserver
    - Frontend
  ```
3. Values are searched in a case insensitive manner such that
  `'webserver' == 'Webserver'`

Closes #418

Co-authored-by: Daniel Carabas <daniel.carabas@puppet.com>

### DIFF
--- a/lib/puppet_litmus/inventory_manipulation.rb
+++ b/lib/puppet_litmus/inventory_manipulation.rb
@@ -53,6 +53,85 @@ module PuppetLitmus::InventoryManipulation
     end
   end
 
+  # Recursively find and iterate over the groups in an inventory. If no block is passed
+  # to the function then only the name of the group is returned. If a block is passed
+  # then the block is executed against each group and the value of the block is returned.
+  #
+  # @param inventory_hash [Hash] Inventory hash from inventory.yaml
+  # @param block [Block] Block to execute against each node
+  def groups_in_inventory(inventory_hash, &block)
+    inventory_hash['groups'].flat_map do |group|
+      output_collector = []
+      output_collector << if block_given?
+                            yield group
+                          else
+                            group['name'].downcase
+                          end
+      output_collector << groups_in_inventory({ 'groups' => group['groups'] }, &block) if group.key? 'groups'
+      output_collector.flatten.compact
+    end
+  end
+
+  # Iterate over all targets in an inventory. If no block is given to the function
+  # it will return the name of every target in the inventory. If a block is passed
+  # it will execute the block on each target and return the value of the block.
+  #
+  # @param inventory_hash [Hash] Inventory hash from inventory.yaml
+  # @param block [Block] Block to execute against each node
+  def targets_in_inventory(inventory_hash)
+    groups_in_inventory(inventory_hash) do |group|
+      if group.key? 'targets'
+        group['targets'].map do |target|
+          if block_given?
+            (yield target)
+          else
+            target['uri'].downcase
+          end
+        end
+      end
+    end
+  end
+
+  # Find all targets in an inventory that have a role. The roles for a target are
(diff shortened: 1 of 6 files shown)


RandomNoun7:
I have modified two modules to work with these modifications to validate that they work. You can find those branches in the following two links:
https://github.com/RandomNoun7/puppetlabs-splunk_hec/tree/MAINT-convert-to-litmus-targets
https://github.com/RandomNoun7/puppetlabs-common_events/tree/MAINT-convert-to-targetted-litmus

I need to fix the existing tests and add new ones, but this code is in a state where it functions to do testing and start conversion on other modules to see how it works.

RandomNoun7:
Worth noting, the acceptance suite passes now. It’s just Rubocop violations causing the failures now.

carabasdaniel:
Hi @RandomNoun7,

This draft looks great. I put up a commit with a fix for the rubocop errors, however we still need to test this out to ensure backwards compatibility (motd is currently failing on appveyor) and we will need to update the official documentation [here](https://github.com/puppetlabs/litmus) to reflect this change before a merge.

RandomNoun7:
I see the Appveyor error. I'll start a Windows instance tomorrow have a look since it does look like one of my new functions is the problem. 

Thanks looking at the rubocop errors for me!

RandomNoun7:
Ok, tests green now. I'll get those docs updates as soon as I can.

carabasdaniel:
@RandomNoun7 Awesome work on this. I'll get this merged as a bugfix and we can release it once the docs are updated.

carabasdaniel (approved):

sheenaajay (approved):

sheenaajay:
@RandomNoun7 Great work. Thank you.

RandomNoun7:
@sheenaajay @carabasdaniel I'm noticing that some of the tests this is run against and some of the inventory code in here makes use of `features` in some targets properties. Does this targeting feature need to be extended to allow targeting features as well as groups and roles?

carabasdaniel:
Hi @RandomNoun7, that should be added in another PR as an extension to this functionality. I'll merge this one for now. Thanks again for the contribution

PR #430, opened 2021-09-08
Remove support for rhel docker images that no longer exist

ZloeSabo:
I've removed the RedHat-7/8 support from the code as the non-namespaced `rhel-7` and `rhel-8` images are no longer available on the docker hub and I wasn't able to find any functioning alternatives to replace them.

I've tried the [official UBI images](https://hub.docker.com/u/redhat), but currently `puppetlabs/provision` [does not have support](https://github.com/puppetlabs/provision/blob/main/tasks/docker.rb#L20) for `rhel` distro that UBI reports. And even if I fork it and add it, RedHat [objects adding OpenSSH server to their repos](https://bugzilla.redhat.com/show_bug.cgi?id=1750907) for UBI, so litmus won't be able to connect there. As for the non-UBI images, I didn't find one that is both maintained and has systemd set up properly so that they could run services.

Without this change, litmus fails with something like

```
localhost: {"_error"=>{"kind"=>"provision/docker_failure", "msg"=>"Attempted to run\ncommand:'docker run -d -it --privileged  --tmpfs /tmp:exec -p 2222:22 --name rhel-7-2222 rhel-7'\nstdout:\nstderr:Unable to find image 'rhel-7:latest' locally\ndocker: Error response from daemon: pull access denied for rhel-7, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.\nSee 'docker run --help'.\n", "backtrace"=>["/tmp/4d3453a8-0565-43e7-80a1-c9923604b48d/provision/lib/task_helper.rb:21:in `run_local_command'", "/tmp/4d3453a8-0565-43e7-80a1-c9923604b48d/provision/tasks/docker.rb:156:in `provision'", "/tmp/4d3453a8-0565-43e7-80a1-c9923604b48d/provision/tasks/docker.rb:198:in `<main>'"], "details"=>{}}}}
```

I would gladly update this PR if someone points me at images that I can use instead.

### DIFF
--- a/exe/matrix_from_metadata
+++ b/exe/matrix_from_metadata
@@ -6,8 +6,6 @@
 require 'json'
 
 IMAGE_TABLE = {
-  'RedHat-7' => 'rhel-7',
-  'RedHat-8' => 'rhel-8',
   'SLES-12' => 'sles-12',
   'SLES-15' => 'sles-15',
   'Windows-2012 R2' => 'windows-2012-r2-core',
--- a/exe/matrix_from_metadata_v2
+++ b/exe/matrix_from_metadata_v2
@@ -6,8 +6,6 @@
 require 'json'
 
 IMAGE_TABLE = {
-  'RedHat-7' => 'rhel-7',
-  'RedHat-8' => 'rhel-8',
   'SLES-12' => 'sles-12',
   'SLES-15' => 'sles-15',
   'Windows-2012 R2' => 'windows-2012-r2-core',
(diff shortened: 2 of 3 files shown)


carabasdaniel:
Hi @ZloeSabo, the images described in the  matrix_from_metadata are used to generate the setup test matrix we use across the IAC supported modules. As you probably seen on most of our modules we use actual VMs to run tests for RHEL, Windows and SLES images - [example run](https://github.com/puppetlabs/puppetlabs-apache/runs/3585893874?check_suite_focus=true). The containers we spin up are managed by the DOCKER_PLATFORMS  definition in this file where possible.

PR #456, opened 2022-08-09
Update bolt version requirement

chelnak:


pmcmaw:
I have tested on a local set up and can confirm this change results in puppet_litmus pulling in the latest version of bolt. Previously it was pulling in v3.23.1 it now pulls v3.25.0

PR #462, opened 2022-10-27
(MAINT) Remove Windows Server 2012R2

chelnak:
Agent installation has been failing on Windows Server 2012R2 for some time now.

It turns out that this is being caused by a mismatch in ciphers between the provisioned boxes and the service that provides nightlies.puppet.com.

The nightlies archive requires a specific set of ciphers which are not supported on the base installs of Server 2012 R2.

This commit removes Server 2012 R2 from the list of images that are used to build the test matrix.


GSPatton (approved):

david22swan:
Technically this would be backwards incompatible. But since this is still a `0.` release module we can set it as a bugfix.

Issue #503, opened 2023-06-22
Unreleased modules needs a manual upload to docker


JGodin-C2C:
## Describe the Bug
When creating a new module or a private module, if the module is not on the forge, one needs to upload it to the docker where the tests are performed

## Expected Behavior
When creating a new module or a private module, litmus should build the local module and upload it to the docker where the test are performed

## Steps to Reproduce
1. Create a new module `pdk new module mymodule`
2. add some dependencies
In metadata.json:
```
    {
      "name": "puppetlabs/stdlib",
      "version_requirement": "=9.0.0"
    },
    {
      "name": "puppetlabs/apache",
      "version_requirement": "=10.1.0"
    }
```
in .fixtures.yml
```
---
fixtures:
  repositories:
    facts: 'https://github.com/puppetlabs/puppetlabs-facts.git'
    puppet_agent: 'https://github.com/puppetlabs/puppetlabs-puppet_agent.git'
    provision: 'https://github.com/puppetlabs/provision.git'
    concat: "https://github.com/puppetlabs/puppetlabs-concat.git"
    portage: "https://github.com/gentoo/puppet-portage.git"
    stdlib: "https://github.com/puppetlabs/puppetlabs-stdlib.git"
    apache: "https://github.com/puppetlabs/puppetlabs-apache.git"
    yumrepo_core:
      repo: "https://github.com/puppetlabs/puppetlabs-yumrepo_core.git"
      puppet_version: '>= 6.0.0'

```
settup the provision.yaml
```
---
default:
  provisioner: docker_exp
  images:
  - litmusimage/centos:stream8
```
4. Launch the litmus  provisioner `pdk bundle exec rake 'litmus:provision[docker_exp, litmusimage/centos:stream8]'`
5. Install the puppet agent `pdk bundle exec rake 'litmus:install_agent'`
6. Try to install the module : `pdk bundle exec rake 'litmus:install_module'`

PR #505, opened 2023-06-29
feat: install an offline module using the tar.gz

JGodin-C2C:
## Summary

This pull request allows the test of a module that have not been released to the forge yet or simply needs to remain private. Please see #503 

## How

The module is built , then uploaded and the installation is done without its dependencies.
dependencies are then parsed using the metadata.json file and installed automatically

## Additional Context
Add any additional context about the problem here. 
- [ ] The puppet agent tries to contact the forge to install the unreleased package. 

## Related Issues (if any)
#503 

## Checklist
- [ ] 🟢 Tested by hand. Modules is installed and dependencies as well.

### DIFF
--- a/lib/puppet_litmus/rake_helper.rb
+++ b/lib/puppet_litmus/rake_helper.rb
@@ -224,15 +224,26 @@ module PuppetLitmus::RakeHelper
     include ::BoltSpec::Run
 
     target_nodes = find_targets(inventory_hash, target_node_name)
-    bolt_result = upload_file(module_tar, File.basename(module_tar), target_nodes, options: {}, config: nil, inventory: inventory_hash.clone)
+    bolt_result = upload_file(module_tar, "/tmp/#{File.basename(module_tar)}", target_nodes, options: {}, config: nil, inventory: inventory_hash.clone)
     raise_bolt_errors(bolt_result, 'Failed to upload module.')
 
     module_repository_opts = "--module_repository '#{module_repository}'" unless module_repository.nil?
-    install_module_command = "puppet module install #{module_repository_opts} #{File.basename(module_tar)}"
+    install_module_command = "puppet module install #{module_repository_opts} /tmp/#{File.basename(module_tar)}"
     install_module_command += ' --ignore-dependencies --force' if ignore_dependencies.to_s.casecmp('true').zero?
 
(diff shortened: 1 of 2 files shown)


jordanbreen28:
Nice one @JGodin-C2C!!
Rubocop seems to be complaining, can you take a look and I'll get to testing this out? :)

lib/puppet_litmus/rake_helper.rb
```diff
@@ -224,15 +224,26 @@ def install_module(inventory_hash, target_node_name, module_tar, module_reposito
     include ::BoltSpec::Run
 
     target_nodes = find_targets(inventory_hash, target_node_name)
-    bolt_result = upload_file(module_tar, File.basename(module_tar), target_nodes, options: {}, config: nil, inventory: inventory_hash.clone)
+    bolt_result = upload_file(module_tar, "/tmp/#{File.basename(module_tar)}", target_nodes, options: {}, config: nil, inventory: inventory_hash.clone)
```
jordanbreen28, line 227: Curious on the choice to change the upload destination, is this required?

lib/puppet_litmus/rake_helper.rb
```diff
@@ -224,15 +224,26 @@ def install_module(inventory_hash, target_node_name, module_tar, module_reposito
...
     raise_bolt_errors(bolt_result, "Installation of package #{File.basename(module_tar)} failed.")
+
+    return bolt_result if metadata_module_dependencies.nil?
+
+    metadata_module_dependencies.each do |dependency|
+      module_repository_opts = "--module_repository '#{module_repository}'" unless module_repository.nil?
+      install_module_command = "puppet module install #{module_repository_opts} #{dependency['name']}"
+      install_module_command += " --version '#{dependency['version_requirement']}'" if dependency.key?('version_requirement')
+
+      bolt_result = run_command(install_module_command, target_nodes, config: nil, inventory: inventory_hash.clone)
```
jordanbreen28, line 244: the module dependencies are installed along with the module using the original `install_module` command, so what's the reasoning behind adding this?

lib/puppet_litmus/rake_helper.rb
JGodin-C2C: I will test without this, i think this is working without this change.

PR #516, opened 2023-07-28
(CAT-1241) - Adding retry when provision failed with Timeout

Ramesh7:
## Summary
We have seen many times the PR CI is gets failed due to Connectivity check Timeout which lead to failure for the workflow, but the same can be successful if we retries.
So adding 3 retries whenever there is timeout failure.

## Additional Context
- N/A

## Related Issues (if any)
- N/A

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.

### DIFF
--- a/lib/puppet_litmus/rake_tasks.rb
+++ b/lib/puppet_litmus/rake_tasks.rb
@@ -99,7 +99,7 @@ namespace :litmus do
     rescue PuppetLitmus::RakeHelper::LitmusTimeoutError
       current_retry_count += 1
       Rake::Task['litmus:tear_down'].invoke(target_names.first)
-      raise if current_retry_count >= retry_count
+      raise if current_retry_count > retry_count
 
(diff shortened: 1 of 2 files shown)


lib/puppet_litmus/rake_tasks.rb
```diff
@@ -94,6 +96,13 @@
           check_connectivity?(inventory_hash_from_inventory_file, target)
         end
       end
+    rescue PuppetLitmus::RakeHelper::LitmusTimeoutError
+      current_retry_count += 1
+      Rake::Task['litmus:tear_down'].invoke(target_names.first)
+      raise if current_retry_count >= retry_count
```
jordanbreen28, line 102: should this be the below to allow for 3 retries?
```suggestion
      raise if current_retry_count > retry_count
```

jordanbreen28 (approved):

PR #549, opened 2024-02-13
(feature) matrix from metadata v3

h0tw1r3:
another matrix generator with a focus on flexibility.

- `matrix.json` is the configuration for supported provisioners and collections.
- runners are now part of the matrix, the key `github_runners` is a hash of provisioners, each with a hash of regex to runner name maps.
- disable provision_service if not in the puppetlabs org
- special hidden flag `--puppetlabs` to enable the `provision_service` even if the repo is not in the puppetlabs org


LukasAud:
Awaiting docs update before merging.

h0tw1r3:
@puppetlabs/devx docs complete, ready for merge

PR #556, opened 2024-04-26
drop Debian 9 stretch from matrix

h0tw1r3:
EOL, vendor repos do not work, and Puppet hasn't built agents for a long time


jordanbreen28:
@h0tw1r3 fix for the tests has went into main - can you rebase? :)

PR #557, opened 2024-04-27
(maint) fix matrix tests broken by bacd4d7

h0tw1r3:


jordanbreen28 (approved):
Thanks @h0tw1r3! It's brought to our attention that the spec tests are exiting 0 on failure.. strange, we will look into this

PR #559, opened 2024-04-30
use dockercli specinfra backend for docker_nodes

h0tw1r3:
This fixes running Serverspec tests against nodes created by the docker_exp provision task. Unlike the docker bolt transport, the Serverspec docker backend uses the docker API. Unfortunately it lacks support for modern docker contexts (notably ssh).

_Only applies to nodes created with the docker_exp provision task_


david22swan (commented):
LGTM

david22swan (approved):

PR #563, opened 2024-05-02
automatically filter provision_service from matrix

h0tw1r3:
When testing changes with workflows on a personal fork, acceptance tests always fail because provision_service hosts are always included. This PR automatically filters them if the owner isn't `puppetlabs`, and the GITHUB_REPOSITORY_OWNER variable is set.


david22swan (approved):
LGTM

Thanks for the work :)

PR #576, opened 2024-10-24
(FEAT) Add PE to matrix_from_metadata_v3

coreymbe:
## Summary
This commit adds a feature to the `matrix_from_metadata_v3`  tool allowing users to pass the option `--pe-include` which will add Puppet Enterprise LTS versions to the collection matrix.

This is done by utilizing the `puppet_forge` gem to make a connection to the Forge API and retrieve a list of PE versions. 

The code in this commit will only add the LTS versions of PE to the collection matrix if the tool is passed the aforementioned option and the puppet version in the LTS falls within the supported range found in the metadata.

**Note**: Ideally this would add LTS and STS to the collection matrix. Once PE 2025.y is released, this could be easily updated to add the STS to the collection as well.

## Additional Context
My team is currently unable to utilize this tool to test our modules against PE. As such we are building our own matrix in separate workflows for PE and OSP. It would be nice to have a way to dynamically determine the LTS and STS PE versions and add them to the collection matrix.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [X] Manually verified.

```
# bundle exec rspec spec/exe/matrix_from_metadata_v3_spec.rb --format documentation

matrix_from_metadata_v3
  without arguments
    run successfully
    generates the matrix
  with puppetlabs GITHUB_REPOSITORY_OWNER
    run successfully
    generates the matrix
  with argument --puppetlabs
    run successfully
    generates the matrix
  with --exclude-platforms "ubuntu-18.04"
    run successfully
    generates the matrix without excluded platforms
  with --platform-exclude "ubuntu-(18.04|22.04)" --platform-exclude "redhat-[89]"
    run successfully
    generates the matrix without excluded platforms
  with --pe-include
    run successfully
    generates the matrix with PE LTS versions

Finished in 3.96 seconds (files took 0.26874 seconds to load)
12 examples, 0 failures
```

### DIFF
--- a/exe/matrix_from_metadata_v3
+++ b/exe/matrix_from_metadata_v3
@@ -147,6 +147,8 @@ begin
 
     opt.on('--runner NAME', String, "Default Github action runner (default: #{default_options[:runner]})") { |o| options.runner = o }
 
+    opt.on('--pe-include', TrueClass, 'Include Puppet Enterprise LTS') { |o| options.pe_include = o }
+
     opt.on('--puppet-include MAJOR', Integer, 'Select puppet major version') { |o| options.puppet_include << o }
     opt.on('--puppet-exclude MAJOR', Integer, 'Filter puppet major version') { |o| options.puppet_exclude << o }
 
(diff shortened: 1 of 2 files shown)


david22swan (approved):

david22swan:
Hey, gonna approve and merge this, but going forward with PRs could you open them from the main fork instead of your own as otherwise the mend scan will not work.

PR #591, opened 2025-07-22
(PA-7608) Use artifactory URL for puppetcore nightlies

skyamgarp:
This change is to use artifactory instead of nightlies


gavindidrichsen (approved):
Thank you!!

PR #627, opened 2026-07-21
(MODULES-11700) Add --collection-platform-exclude to matrix_from_metadata_v3

imaqsood:
# Add `--collection-platform-exclude` to matrix_from_metadata_v3

## Problem

`matrix_from_metadata_v3` emits `matrix.platforms` and `matrix.collection` as two
independent lists. The consuming GitHub Actions workflow uses them as a
`strategy.matrix`, i.e. a full cartesian product — every platform is tested
against every collection.

That breaks down when a platform is valid for one Puppet major but not another.
Example: Ubuntu 20.04 (focal) is supported by Puppet 8 but dropped by Puppet 9,
so `ubuntu-20.04 × puppetcore9-nightly` fails at agent install (404 — no focal
package for Puppet 9). Today the only levers are:

- `--platform-exclude` — removes the platform from **all** collections (would also
  drop the valid Puppet 8 coverage), and
- `--puppet-exclude` — removes a Puppet major from **all** platforms.

Neither can drop a single platform×collection cell.

## Change

Add a repeatable `--collection-platform-exclude MAJOR:PLATFORM_REGEX` flag. It emits
a GitHub Actions matrix `exclude:` entry for each matching platform×collection
combination, leaving both the platform and the collection available for every other
combination.

```
# ubuntu-20.04 stays in the Puppet 8 lane, dropped only from Puppet 9
matrix_from_metadata_v3 --nightly --collection-platform-exclude 9:ubuntu-20.04
```

`MAJOR` is the Puppet major (integer, matched against the collection, so it applies
to nightly and non-nightly alike). `PLATFORM_REGEX` is matched case-insensitively
against the platform label. Invalid specs (missing `MAJOR:` or non-numeric major)
raise an `OptionParser::InvalidArgument`.

No behaviour change unless the flag is supplied.

## Notes / review points

- The `exclude` entry references the full `platforms` object and the `collection`
  value verbatim from the generated lists, so GitHub Actions matches them by
  deep-equality. This relies on GHA's `exclude` supporting object-valued matrix
  dimensions (the workflow already treats `platforms`/`collection` as objects).
  Worth a smoke-test in a real run; if it proves unreliable, the alternative is to
  emit a single `include:` list instead of two independent dimensions.
- Alternative considered: deriving the exclusions automatically from AIO
  support data (which Puppet major ships which OS). That needs a data source the
  tool doesn't currently carry; the explicit flag is minimal and mirrors the
  existing `--platform-exclude` idiom. Auto-derivation can layer on later.

## Tests

Adds examples to `spec/exe/matrix_from_metadata_v3_spec.rb` covering: the emitted
`exclude` for the matching pair, the platform/collection remaining available
elsewhere, the `::notice` log line, and the invalid-spec error path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ in the project module root directory run `bundle exec matrix_from_metadata_v3`
 | --puppet-exclude    | MAJOR |                   | Filter puppet major version |
 | --platform-include  | REGEX |                   | Select platform |
 | --platform-exclude  | REGEX |                   | Filter platform |
+| --collection-platform-exclude | MAJOR:REGEX | | Filter a platform for one puppet major only, emitted as a GitHub Actions matrix exclude so the platform stays in other collections. Repeatable |
 | --arch-include      | REGEX |                   | Select architecture |
 | --arch-exclude      | REGEX |                   | Filter architecture |
 | --provision-prefer  | NAME  | docker            | Prefer provisioner |
@@ -87,6 +88,10 @@ in the project module root directory run `bundle exec matrix_from_metadata_v3`
 * Exclude platforms
   ```sh
   matrix_from_metadata_v3 --platform-exclude redhat-7 --platform-exclude ubuntu-18.04
+  ```
+* Exclude a platform for one puppet major only (kept in other collections). E.g. ubuntu-20.04 has no Puppet 9 agent but is valid on Puppet 8:
+  ```sh
+  matrix_from_metadata_v3 --nightly --collection-platform-exclude 9:ubuntu-20.04
   ```
 * Exclude architecture
   ```sh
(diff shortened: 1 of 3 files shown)


david22swan (changes requested):
If you could update the Readme to include a reference to this and resolve the copilot comments that would be good: https://github.com/puppetlabs/puppet_litmus#matrix_from_metadata_v3

Otherwise LGTM and a valid use case

imaqsood:
Thanks @david22swan! Addressed in a1bc90c:

- **README** — added `--collection-platform-exclude` to the options table and an example under `## matrix_from_metadata_v3`.
- **Copilot 1 (invalid regex)** — an invalid `PLATFORM_REGEX` (e.g. `9:[`) now raises a clean `OptionParser::InvalidArgument` instead of a `RegexpError` stack trace.
- **Copilot 2 (spec exactness)** — the test now parses the matrix JSON and asserts `exclude` is *exactly* the one expected `platforms`×`collection` entry (via `contain_exactly`), so stray entries would fail it.

david22swan (approved):

